PR #13, opened 2023-05-31
Schema browser refresh on connection

shamby87:


src/views/schemaBrowser.ts
```diff
@@ -123,18 +123,25 @@ export default class schemaBrowser {
...
+            await getInstance().setConfig(config);
           }
 
           vscode.window.showInformationMessage(`Current schema set to ${schema}.`);
         }
       })
     )
+
+    getInstance().onEvent(`connected`, () => {
+      fetchSystemInfo().then(() => {
```
worksofliam, line 140: The call to `fetchSystemInfo` is not required here.

worksofliam (changes requested):
Check out my comment. Should be an easy change, as it is not required here.

src/views/schemaBrowser.ts
shamby87: Fixed

shamby87:
Yep, I was going to try that and forgot. Should be good now

PR #18, opened 2023-06-23
Add right-click options to Schema Browser items

shamby87:


src/views/schemaBrowser.ts
```diff
@@ -108,6 +108,201 @@ export default class schemaBrowser {
         }
       }),
 
+      vscode.commands.registerCommand(`vscode-db2i.getRelatedObjects`, async (object: SQLObject) => {
+        if (object) {
+          try {
+            const content = `SELECT SQL_OBJECT_TYPE, SCHEMA_NAME, SQL_NAME, LIBRARY_NAME,
```
worksofliam, line 114: This query belongs in an API instead of embedding the query in the command.

src/views/schemaBrowser.ts
```diff
@@ -108,6 +108,201 @@ export default class schemaBrowser {
...
+            vscode.window.showErrorMessage(e);
+          }
+        }
+      }),
+
+      vscode.commands.registerCommand(`vscode-db2i.getIndexes`, async (object: SQLObject) => {
+        if (object) {
+          try {
+            // Maybe choose/rename which columns to get?
+            const content = `SELECT * FROM QSYS2.SYSINDEXSTAT WHERE TABLE_SCHEMA = '${object.schema}' and TABLE_NAME = '${object.name}'`;
```
worksofliam, line 132: This query belongs in an API instead of embedding the query in the command.

src/views/schemaBrowser.ts
```diff
@@ -108,6 +108,201 @@ export default class schemaBrowser {
...
+            vscode.window.showErrorMessage(e);
+          }
+        }
+      }),
+
+      vscode.commands.registerCommand(`vscode-db2i.advisedIndexes`, async (object: SQLObject) => {
+        if (object) {
+          try {
+            // Maybe choose/rename which columns to get?
+            const content = `SELECT * FROM QSYS2.SYSIXADV WHERE TABLE_SCHEMA = '${object.schema}' and TABLE_NAME = '${object.name}'`;
```
worksofliam, line 148: This query belongs in an API instead of embedding the query in the command.

src/views/schemaBrowser.ts
```diff
@@ -108,6 +108,201 @@ export default class schemaBrowser {
...
+
+      vscode.commands.registerCommand(`vscode-db2i.clearAdvisedIndexes`, async (object: SQLObject) => {
+        if (object) {
+          const result = await vscode.window.showWarningMessage(`Are you sure you want to clear all of the advised index rows from the Index Advisor for ${object.name}?`,  {
+            modal: true,
+          }, 'No', 'Yes');
+          
+          if(result === 'Yes') {
+            try {
+              const query = `DELETE FROM QSYS2.SYSIXADV WHERE TABLE_SCHEMA = '${object.schema}' and TABLE_NAME = '${object.name}'`;
```
worksofliam, line 168: This query belongs in an API instead of embedding the query in the command.

worksofliam (changes requested):
@shamby87 A good change, but I think the code could be a little cleaner.

I am not a big fan of embedding queries into commands. If we can turn them into their own methods we can re-use them. The `database` directory has some database queries wrapped into methods and I'd like to see more added instead of via this PR.

Also, uou are you using two ways to run a query, but I think it's worth standardising on one (I prefer the `runSQL` over the command)

* https://github.com/halcyon-tech/vscode-db2i/pull/18/files#diff-87342e72eeaa1ab60a5dad8db3c4d2464e1fa66043961d135588854c0e168ab2R169
* https://github.com/halcyon-tech/vscode-db2i/pull/18/files#diff-87342e72eeaa1ab60a5dad8db3c4d2464e1fa66043961d135588854c0e168ab2R149

Feel free to reach out to me and we can work on this together some.

worksofliam (changes requested):
Change to using APIs is almost perfect. Last part that needs some cleaning up is removing any 'ui' code (`showXMessage`) from the API methods and instead letting the commands/UI code handle them.

worksofliam (approved):
Code looks good to me. @davecharron over to you for some testing.

davecharron:
### Rename action - Bug

Cancelling the rename prompt actually performs the rename action using the name 'undefinded'

davecharron:
### Clear action - Comments

1. There is no confirmation message after clicking Yes to clear the data.
2. Do we really need No/Yes/Cancel on the confirmation?  No and Cancel would seem to do exactly the same thing.

davecharron:
# Testing Feedback...  In progress, actively updating this comment

## General
ℹ️ Do we really need No/Yes/Cancel on the confirmation dialogs?  No and Cancel would seem to do exactly the same thing. 

## Clear action
⚠️ There is no confirmation message after clicking Yes to clear the data.
⚠️ The failure message when the object does not exist or the clear fails because the object is locked or user doesn't have authority to it, takes a long time ( 30 seconds? ) to display, if it does at all.  
If the API we use to send the command to the system allows job level control it would be good to change the default wait time to something really small like 1 rather than the default 30 so the feedback is more timely.

## Copy data
ℹ️ I would suggest that Data be capitalized.  This comment really applies to all extension development in that the capitalization of menu items should be consistent.
⚠️ The custom UI does not allow for specifying a member name and label.
⚠️ There is no confirmation message after clicking Copy to copy the data.

## Delete
⚠️ Same comment as that for the Clear action with regard to failure messages.

## Rename action
🛑 Cancelling the rename prompt actually performs the rename action using the name 'undefinded'
⚠️ The action should really be issuing an [SQL RENAME](https://www.ibm.com/docs/en/i/7.5?topic=statements-rename) rather than the RNMOBJ CL command.  This would also allow for setting both the SQL and system identifier for the object.

shamby87:
> Do we really need No/Yes/Cancel on the confirmation dialogs? No and Cancel would seem to do exactly the same thing.

- This is a good question and it is something I didn't like. For the popup message, it requires that there be a default "nothing" action for if a user exits the window, which is the `cancel` button. I can set the `no` button to replace the `cancel` so there would just be 2 buttons, but then I have no way to make `no` the default answer like we discussed. Having it this way allows `no` to be the default. @worksofliam do you have any way around this? [Here](https://github.com/halcyon-tech/vscode-db2i/pull/18/files#diff-87342e72eeaa1ab60a5dad8db3c4d2464e1fa66043961d135588854c0e168ab2R160-R162) is an example of where I do this.


> There is no confirmation message after clicking Yes to clear the data.
> There is no confirmation message after clicking Copy to copy the data.

- Confirmation as in "Success the data was cleared" or another "Yes I really want to do this"?


> The failure message when the object does not exist or the clear fails because the object is locked or user doesn't have authority to it, takes a long time ( 30 seconds? ) to display, if it does at all.
If the API we use to send the command to the system allows job level control it would be good to change the default wait time to something really small like 1 rather than the default 30 so the feedback is more timely.

- @worksofliam I used `getInstance().getConnection().runCommand()`, is there a way to control the wait time for a response here? Is there an environment variable that can control this?


I'll also look into fixing the other issues, thank you for all the feedback!

worksofliam:
@shamby87 There is no wait time support, but you should look into the `withProgress` VS Code API to help show the user something is running in the background.

package.json
```diff
@@ -273,6 +313,46 @@
...
+          "group": "db2data@1"
+        },
+        {
+          "command": "vscode-db2i.copyData",
+          "when": "viewItem == table || viewItem == view",
+          "group": "db2data@2"
+        },
+        {
+          "command": "vscode-db2i.deleteObject",
+          "when": "viewItem == table || viewItem == view || viewItem == alias || viewItem == constraint || viewItem == function || viewItem == variable || viewItem == index || viewItem == procedure || viewItem == sequence || viewItem == package || viewItem == type",
```
davecharron, line 338: Missing support for deleting triggers

src/database/schemas.ts
```diff
@@ -142,4 +143,22 @@ export default class Database {
 
     return formatted;
   }
+  
+  static async deleteObject(objectPath: string, type: string): Promise<void> {
```
davecharron, line 147: Instead of objectPath, do we want to maintain the pattern of specifying schema and object name as separate values?

src/database/schemas.ts
```diff
@@ -142,4 +143,22 @@ export default class Database {
...
+    const query = `DROP ${type} IF EXISTS ${objectPath}`;
+    await getInstance().getContent().runSQL(query);
+  }
+  
+  static async renameObject(schema: string, oldName: string, newName: string): Promise<void> {
+    const command = `RNMOBJ OBJ(${schema}/${oldName}) OBJTYPE(*FILE) NEWOBJ(${newName})`;
```
davecharron, line 153: Use the RENAME SQL statement instead of RNMOBJ

src/database/table.ts
```diff
@@ -36,4 +38,67 @@ export default class Table {
 
     return content.runSQL(sql);
   }
+
+  static getRelatedObjects(schema: string, name: string): void {
+    const content = `SELECT SQL_OBJECT_TYPE, SCHEMA_NAME, SQL_NAME, LIBRARY_NAME,
+              SYSTEM_NAME, OBJECT_OWNER, LONG_COMMENT, OBJECT_TEXT, LAST_ALTERED 
+              FROM TABLE(SYSTOOLS.RELATED_OBJECTS('${schema}', '${name}')) ORDER BY SQL_NAME`;
```
davecharron, line 45: I would suggest the following column order:
```
SQL_NAME, SYSTEM_NAME, SCHEMA_NAME, LIBRARY_NAME, SQL_OBJECT_TYPE, 
              OBJECT_OWNER, LAST_ALTERED, OBJECT_TEXT, LONG_COMMENT
```
Would also order by SQL_OBJECT_TYPE then by SQL_NAME

src/database/table.ts
```diff
@@ -36,4 +38,67 @@ export default class Table {
...
+    vscode.commands.executeCommand(`vscode-db2i.runEditorStatement`, {
+      content,
+      type: `statement`,
+      open: false,
+    });
+  }
+
+  static getAdvisedIndexes(schema: string, name: string): void {
+    // Maybe choose/rename which columns to get?
+    const content = `SELECT * FROM QSYS2.SYSIXADV WHERE TABLE_SCHEMA = '${schema}' and TABLE_NAME = '${name}'`;
```
davecharron, line 66: Should order the results by TIMES_ADVISED descending.

src/views/schemaBrowser.ts
```diff
@@ -103,7 +104,154 @@ export default class schemaBrowser {
...
+      
+      vscode.commands.registerCommand(`vscode-db2i.clearData`, async (object: SQLObject) => {
+        if (object) {
+          const result = await vscode.window.showWarningMessage(`Are you sure you want to clear ${object.name}?`, {
+            modal: true,
+          }, 'No', 'Yes');
+
+          if(result === 'Yes') {
+            try {
+              Table.clearFile(object.schema, object.name);
```
davecharron, line 207: I think this might need to use `await`

src/views/schemaBrowser.ts
```diff
@@ -103,7 +104,154 @@ export default class schemaBrowser {
...
+              {text: '*NO', description: '*NO', value: '*NO'},
+              {text: '*YES', description: '*YES', value: '*YES'},
+            ])
+            .addButtons(
+              {id: 'copy', label:'Copy'},
+              {id: 'cancel', label:'Cancel'}
+            )
+            .loadPage<any>('Copy File');
+          
+          if(page && page.data) {
```
davecharron, line 237: > SonarLint suggests: Prefer using an optional chain expression instead, as it's more concise and easier to read.sonarlint(typescript:S6582)
`page?.data`

src/views/schemaBrowser.ts
```diff
@@ -103,7 +104,154 @@ export default class schemaBrowser {
...
+
+      vscode.commands.registerCommand(`vscode-db2i.clearAdvisedIndexes`, async (object: SQLObject) => {
+        if (object) {
+          const result = await vscode.window.showWarningMessage(`Are you sure you want to clear all of the advised index rows from the Index Advisor for ${object.name}?`,  {
+            modal: true,
+          }, 'No', 'Yes');
+          
+          if(result === 'Yes') {
+            try {
+              Table.clearAdvisedIndexes(object.schema, object.name);
```
davecharron, line 150: I think this might need to use await

davecharron (changes requested):
I've looked through the code and have added some comments.

davecharron (approved):
I have no further comments.  Code looks good.

davecharron:
> > Do we really need No/Yes/Cancel on the confirmation dialogs? No and Cancel would seem to do exactly the same thing.
> 
>     * This is a good question and it is something I didn't like. For the popup message, it requires that there be a default "nothing" action for if a user exits the window, which is the `cancel` button. I can set the `no` button to replace the `cancel` so there would just be 2 buttons, but then I have no way to make `no` the default answer like we discussed. Having it this way allows `no` to be the default.

What do folks think about having just two buttons then?  One that affirms the action (Ex. Delete) along with the Cancel button.  The wording of the messages might necessarily need to change to make sense.

Example:


[Notifications | Visual Studio Code Extension API](https://code.visualstudio.com/api/ux-guidelines/notifications)

davecharron:
Sorry, didn't mean to close the PR.  Was adding a comment and meant to abandon it.

src/views/schemaBrowser.ts
```diff
@@ -103,7 +104,197 @@ export default class schemaBrowser {
...
+            }
+          }
+        }
+      }),
+
+      vscode.commands.registerCommand(`vscode-db2i.renameObject`, async (object: SQLObject) => {
+        if (object) {
+          const name = await vscode.window.showInputBox({
+            title: "New File",
+            prompt: "Enter new name",
```
davecharron, line 193: Pre-populate the rename prompt with the current object name.
`value: object.name`

davecharron (commented):
One more suggestion

src/views/schemaBrowser.ts
```diff
@@ -103,7 +104,197 @@ export default class schemaBrowser {
...
+            ])
+            .addSelect('outFmt', 'Print format', [
+              {text: '*CHAR', description: '*CHAR', value: '*CHAR'},
+              {text: '*HEX', description: '*HEX', value: '*HEX'},
+            ])
+            .addButtons(
+              {id: 'copy', label:'Copy'},
+              {id: 'cancel', label:'Cancel'}
+            )
+            .loadPage<any>('Copy File');
```
davecharron, line 272: Include the qualified object name in the page title.
```suggestion
            .loadPage<any>(`Copy File - ${object.schema}.${object.name}`);
```

worksofliam (approved):
Did a final review include a change to the API code with Brock. Let's merge this!

PR #20, opened 2023-06-27
add history toggle for runEditorStatement

shamby87:


src/views/results/index.ts
```diff
@@ -236,7 +236,7 @@ export function initialise(context: vscode.ExtensionContext) {
                 }
               }
               
-              if (statement.type === `statement`) {
+              if (statement.type === `statement` && statement.history != false) {
```
worksofliam, line 239:
```suggestion
              if (statement.type === `statement` && statement.history !== false) {
```

worksofliam: I would recommend using a strict not equal here for consistency.

worksofliam (changes requested):

worksofliam (approved):
Thanks @shamby87!

PR #24, opened 2023-07-06
Fix/sql loading indicator

worksofliam:
Fixes bug where loading indicator was not appearing when running scrolling queries.


davecharron (approved):
All looks good to me.

PR #28, opened 2023-07-09
Use boolean over "true" and "false" strings

worksofliam:
Uses `boolean` type over string type `"true"|"false"` for JDBC options. Also updated UI to respect true/false values as boolean types.


ajshedivy (commented):
lgtm

PR #3, opened 2022-04-14
Add aggregate functions snippets

novy400:
add all of aggregate functions.

PR #2, opened 2022-03-31
Add scalar functions snippets

novy400:
I hope it's ok.
I have an issue with choices and placeholders in snippets coding.
for Example I Don't know how to implement correctly scalar functions like *TIMESTAMP*
regards
yvon

PR #30, opened 2023-07-10
Initial completion item provider

worksofliam:


src/database/schemas.ts
```diff
@@ -16,7 +16,7 @@ const typeMap = {
 };
 
 export default class Database {
-  static async getObjects(schema: string, type: SQLType, details: PageData = {}): Promise<BasicSQLObject[]> {
+  static async getObjects(schema: string, type: SQLType | string, details: PageData = {}): Promise<BasicSQLObject[]> {
```
worksofliam, line 19: `type` parameter should not have the additional `string` type here. Calling methods should only support `SQLType`

src/language/providers/completionProvider.ts
```diff
@@ -0,0 +1,155 @@
...
+  );
+}
+
+async function getObjectCompletions(
+  curSchema: string,
+  sqlTypes: { [index: string]: any }
+) {
+  const list = [];
+  for (let key in sqlTypes) {
+    let value = sqlTypes[key];
```
worksofliam, line 55: You could cast `value` to be `sqlTypes[key] as SQLType`, but you might not need to after the interface for `completionTypes` is implemented.

src/language/providers/completionProvider.ts
```diff
@@ -0,0 +1,155 @@
...
+async function getTableItems(schema: string, name: string) {
+  const items = await Table.getItems(schema, name);
+  return items.map((i) =>
+    createCompletionItem(i.COLUMN_NAME, CompletionItemKind.Field)
+  );
+}
+
+async function getObjectCompletions(
+  curSchema: string,
+  sqlTypes: { [index: string]: any }
```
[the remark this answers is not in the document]
ajshedivy: Interface created for the `CompletionType`

src/language/providers/completionProvider.ts
```diff
@@ -0,0 +1,187 @@
...
+): CompletionItem {
+  const item = new CompletionItem(name, kind);
+  item.detail = detail;
+  item.documentation = documentation;
+  item.sortText = sortText;
+  return item;
+}
+
+function getColumnAtributes(column: TableColumn): string {
+  return `Field: ${column.COLUMN_NAME}\n Type: ${column.DATA_TYPE}\n HAS_DEFAULT: ${column.HAS_DEFAULT}\n IS_IDENTITY: ${column.IS_IDENTITY}\n IS_NULLABLE: ${column.IS_NULLABLE}`;
```
worksofliam, line 54: Instead of returning a single string with the use of newline (`\n`), let's build an array of lines and then use `.join('\n')` instead.

src/language/providers/completionProvider.ts
```diff
@@ -0,0 +1,187 @@
...
+          table.name,
+          value.icon,
+          `Type: ${value.label}`,
+          `Schema: ${table.schema}`,
+          value.order
+        )
+      );
+    });
+
+    const results = await Promise.all(promises);
```
worksofliam, line 95: Let's change `Promise.all` to use `Promise.allSettled` as that will handle any unfortunate crashes that may occur.

Be sure to rewrite the flat logic beloe because `allSettled` has a different signiture.

src/language/providers/completionProvider.ts
```diff
@@ -0,0 +1,187 @@
...
+    const list = results.flat();
+    completionItemCache.set(curSchema, list);
+  }
+  return completionItemCache.get(curSchema);
+}
+
+async function getCompletionItemsForTriggerDot(
+  objectRefs,
+  currentStatement,
```
worksofliam, line 104: Types are missing from these parameters. Let's add those!

src/language/providers/completionProvider.ts
```diff
@@ -0,0 +1,187 @@
...
+  offset
+): Promise<CompletionItem[]> {
+  let list: CompletionItem[] = [];
+  const curRef = currentStatement.getReferenceByOffset(offset);
+  const curSchema = curRef.object.schema;
```
worksofliam, line 109: Is there a chance that `curRef` can be undefined here? Additional safety logic might be required.

src/language/providers/completionProvider.ts
```diff
@@ -0,0 +1,187 @@
...
+    const objectCompletions = await getObjectCompletions(
+      curSchema,
+      completionTypes
+    );
+    list.push(...objectCompletions);
+  }
+  return list;
+}
+
+async function getCompletionItemsForOtherTriggers(objectRefs) {
```
worksofliam, line 140: I think the name of this function could be shorter or more precise. Perhaps `getCompletionItemsForRefs`.

Also, the parameter could do with a type.

src/language/providers/completionProvider.ts
```diff
@@ -0,0 +1,187 @@
...
+  const promises = objectRefs.map((ref) =>
+    ref.object.name && ref.object.schema
+      ? getTableItems(ref.object.schema, ref.object.name)
+      : Promise.resolve([])
+  );
+
+  const completionItemsArray = await Promise.all(promises);
```
worksofliam, line 147: Also to use `allSettled` here.

src/language/providers/completionProvider.ts
```diff
@@ -0,0 +1,187 @@
...
+
+      // if s is undefined, assume ctrl+ space trigger
+      if (s === undefined) {
+        return getCompletionItemsForOtherTriggers(objectRefs);
+      }
+
+      if (trigger === "." || s.type === `dot`) {
+        const defs = sqlDoc.getDefinitions();
+        return getCompletionItemsForTriggerDot(
+          objectRefs,
```
worksofliam, line 178: Should `objectRefs` be `defs` here? `defs` is marked as undefined.

src/language/providers/completionItemCache.ts
```diff
@@ -0,0 +1,17 @@
+import { CompletionItem } from "vscode";
+import LRU from "lru-cache";
```
worksofliam, line 2: Where does this package come from? `lru-cache` is on npm, but there's no change to the `package.json` file in this PR.

worksofliam: Just ran `npm ls lru-cache` and saw it got installed via another package - awesome. Nonetheless, we still need it as an explicit `dependency` requirement in the `package.json` file.

src/language/providers/completionItemCache.ts
```diff
@@ -0,0 +1,17 @@
...
+
+export let updateCache: Set<string> = new Set<string>();
```
worksofliam, line 4: I think an improved name could be used for more clarity. Perhaps `cacheForUpdate` or `changedCache`. It wasn't clear to me at first what the goal of this Set was.

src/connection/query.ts
```diff
@@ -96,6 +98,15 @@ export class Query<T> {
       throw new Error(queryResult.error || `Failed to run query (unknown error)`);
     }
     this.correlationId = queryResult.id;
+    
+    if (this.sql.toUpperCase().startsWith("CREATE")) {
```
worksofliam, line 102: This logic doesn't belong in this class. Perhaps it would be best suited in the `vscode-db2i.runEditorStatement` command as defined in `src/views/results/index.ts`, after we've parsed the statement and got `StatementInfo`. Perhaps a property could be added to return the references in `parseStatement`

worksofliam (commented):
I have done a few things here:

* Updated the branch to have the latest commits from master (was 130+ behind!) and looks like it is ok.
* Left a review as you can see from my comments. Some types were missing, some Promise logic should change and some simple package stuff.

Let me know if you have any questions. Thanks!

src/connection/query.ts
ajshedivy: The logic behind this is we only want to signal a change in a schema if we see a create statement, AND the operation was successful. Is there a way to detect if an SQL statement is valid before running? 

For instance, if I copy the same logic here to `src/views/results/index.ts`, I will add the current schema to the changedCache even if my create statement didnt work, thus the next time we query that schema, we will not use the cached items.

ajshedivy:
@worksofliam I addressed most comments from your review. The only thing left is the discussion on how/where we should update the completion item cache.

ThePrez:
Use case failure:

```sql 
select * from QIWS.qmenusrc
```
Completion for `qmenusrc` seems to work fine, but then I move the cursor to `*`, delete the `*`, 
```sql
select  from qiws.qmenusrc
```
and press `<ctrl>`+`<space>` 

I get some unexpected completion items in the suggestions
image
(Technically correct but I did not expect to see the special registers suggested unless I was on sysdummy1 or somesuch)

In any event, the special register values are not inserted into the SQL properly. For instance, choosing `current user` results in the following statement:
```sql
selecurrent user  from  qiws.qmenusrc
```

ThePrez:
**JG issue #2**

column assistance doesn't seem to work for partially-filled-in column names. For instance
```sql
select SRC from  qiws.qmenusrc
```

If I place the cursor to immediately after the `SRC` and press `<ctrl>`+`<space>`, I'd expect suggestions for columns starting with `SRC`. However, I get "no suggestions found"

ThePrez:
**JG Issue #3** 

Some suggestions-as-you-type are off, for instance here, where I typed `SRC` and am suggested the `score` function, which doesn't match what I typed. (_Maybe_ this is working as designed as the function suggestions might be intentionally fuzzy-matched ??) 

image

ThePrez:
**JG Issue #4**

Run the following SQL statement, using completion provider to help with `SRCDTA`
```sql
select SRCDTA from  qiws.qmenusrc ORDER BY SRCSEQ
```

press `<backspace>` until the `ORDER BY` clause just has `SRC`, press `<ctrl>`+`<space>`. The only suggestion is `SRCDTA`

image


Certainly related to https://github.com/halcyon-tech/vscode-db2i/pull/30#issuecomment-1716526373 but might also imply something about the underlying cache mechanism so documenting separately

ajshedivy:
> **JG Issue #1:**
> 
> ```sql
> select * from QIWS.qmenusrc
> ```
> 
> Completion for `qmenusrc` seems to work fine, but then I move the cursor to `*`, delete the `*`,
> 
> ```sql
> select  from qiws.qmenusrc
> ```
> 
> and press `<ctrl>`+`<space>`
> 
> I get some unexpected completion items in the suggestions image (Technically correct but I did not expect to see the special registers suggested unless I was on sysdummy1 or somesuch)
> 
> In any event, the special register values are not inserted into the SQL properly. For instance, choosing `current user` results in the following statement:
> 
> ```sql
> selecurrent user  from  qiws.qmenusrc
> ```

There isn't anything in the completion provider that directly references the special registers. There may be a way to suppress these from showing up in the list of completion items. I did notice that the `c` and `t` are highlighted which may suggest that when ctrl+space is pressed, there is an incorrect reference to those characters causing those registers to appear at the top of the list. 
@worksofliam  thoughts?

worksofliam:
@ajshedivy 

> There isn't anything in the completion provider that directly references the special registers.

Correct. But, we have globals and built in defined in the `package.json` file ([here](https://github.com/halcyon-tech/vscode-db2i/blob/main/package.json#L36)) which is what is providing them.

> the special register values are not inserted into the SQL properly

This is likely an issue specific to VS Code more than anything. I will see about recreating this.

worksofliam:
**SF issue #5** 

image

> It didn't show me the column names, should it have

ajshedivy:
@ThePrez 

the rest of the issues all have to do with a known limitation that we only provide completion items based on the `.` trigger or ctrl+space trigger with the cursor on no current references (pressing ctrl+space in middle of column name wont return anything). I think I have a fix for this.

worksofliam:
@forstie @ThePrez The latest build should have fixes for the majority of issues you had.

ThePrez:
**JG issue number 5**

Context assist for `CALL` statements seems to look for views, rather than procedures

image

ThePrez:
**JG issue 6**

suggestions after `schema.` include functions that don't make sense
image

davecharron:
When using system naming, should the / be a content assist trigger in addition to the . and Ctrl+Space?

davecharron:
A nice addition when choosing table/view/alias columns would be to include an "All Columns" completion item that when selected returns a comma delimited list of all the columns from the object.

worksofliam:
I am now considering this PR complete (mostly due to size) and stable. Further features should be tracked in an issue and then into a new PR.

PR #32, opened 2023-07-11
fix: "Display job log" action

ThePrez:


worksofliam (approved):
Good cleanup. No idea why I puit the SQL in the UI code, but this changes that and even adds a test. Thanks!

PR #33, opened 2023-07-11
Remove server component check after first

worksofliam:
I noticed the tests were running slow and it's because we're checking if the server component is installed on every single test.

We only need to check if it's installed in the first test case, then we check the local property after that.

### DIFF
--- a/src/connection/serverComponent.ts
+++ b/src/connection/serverComponent.ts
@@ -16,8 +16,8 @@ const octokit = new Octokit();
 const ExecutablePathDir = `$HOME/.vscode/`;
 
 export class ServerComponent {
+  private static installed: boolean = false;
   static outputChannel: OutputChannel;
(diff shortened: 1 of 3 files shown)


ThePrez:
Would it be better to push logic into `ServerComponent.initialise()` so that it simply knows the install check is completed? Such a change would be safer if the order of the test cases (or functional code) ever get shuffled

worksofliam:
No. I think the purpose of initialise is correct. Though, I do think long term 'installed' should be private and there should be a method that returns a Boolean instead.

ThePrez:
> No. I think the purpose of initialise is correct. Though, I do think long term 'installed' should be private and there should be a method that returns a Boolean instead.

that makes sense. I like `isInstalled()`

ThePrez (approved):

PR #35, opened 2023-07-11
Fix/delim_names

worksofliam:


worksofliam:
Tests have been created for many instances where both system name and SQL name can be used, using the test data in the attached issue.

image

worksofliam:
@davecharron This PR is now ready for review.

davecharron:
I suppose the delimiting quotes do not need to be displayed in the tree, but the case within the name must be preserved.
Consider...
image
``` sql
-- These tables and their primary keys are not the same.  Case within quotes is significant.
CREATE TABLE "TestDelimiters"."Delimited Table" ("Delimited Column" INTEGER DEFAULT NULL, 
CONSTRAINT "TestDelimiters"."Delimited Key" PRIMARY KEY ("Delimited Column"));

CREATE TABLE "TestDelimiters"."delimited table" ("delimited column" INTEGER DEFAULT NULL, 
CONSTRAINT "TestDelimiters"."delimited key" PRIMARY KEY ("delimited column"));

```

src/database/statement.ts
```diff
@@ -8,4 +8,16 @@ export default class Statement {
       linesBetweenQueries: 2, // Defaults to 1
     });
   }
+
+  static delimName(name: string) {
+    if (name.startsWith(`"`) && name.endsWith(`"`)) return name;
+    if (name.includes(` `)) return `"${name}"`;
+    if (name.length <= 10) return name.toUpperCase();
+    else return `"${name}"`;
+  }
```
davecharron, line 17: With a method like this, the calling context is important.
Did the value come from an SQL catalog, or did the user type in the value?
Values from the catalog are already normalized, but user input could be anything.

For example, input value `create_sql_sample`
See test in `testing/database.ts` at line 71

If that value came from a catalog, then the case is very likely significant, and thus must be delimited in order to preserve it.
But if the value was typed in by the user, since they did not delimit the value, it should be folded to uppercase, because as it is it is a valid ordinary identifier which can be up to 128 characters long.

src/testing/database.ts
```diff
@@ -0,0 +1,143 @@
...
+
+      const sqlDelim = Statement.delimName(sqlSchema);
+      assert.strictEqual(sqlDelim, sqlSchema); // No change expect
+
+      const withSpace = Statement.delimName(`my object`);
+      assert.strictEqual(withSpace, `"my object"`);
+
+      const longName = Statement.delimName(`create_sql_sample`);
+      assert.strictEqual(longName, `"create_sql_sample"`);
+    }},
```
davecharron, line 73: See my comment in `database/statement.ts`
We should add comments here to make it clear what we are testing.
This test is expecting the return value be delimited, indicating that the input value would have come from an SQL catalog.

src/database/schemas.ts
```diff
@@ -134,11 +135,14 @@ export default class Database {
   static async generateSQL(schema: string, object: string, type: SQLType): Promise<string> {
     const content = instance.getContent();
```
davecharron, line 136:
```suggestion
```
Not used

src/database/schemas.ts
```diff
@@ -134,11 +135,14 @@ export default class Database {
...
 
-    schema = schema.toUpperCase();
+    schema = Statement.noQuotes(Statement.delimName(schema));
+    object = Statement.noQuotes(Statement.delimName(object));
+
+    // Remove plural and convert to uppercase. Needs work
+    let validType: string = type.substring(0, type.length-1).toUpperCase();
```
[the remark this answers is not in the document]
worksofliam: Replaced with type map in new commit.

src/database/statement.ts
worksofliam: > Did the value come from an SQL catalog, or did the user type in the value?

Yeah, I realized half-way through this impl. that I might need to do something like this. I am going to add an additional parameter to handle these cases.

src/testing/database.ts
worksofliam: Test updated.

worksofliam:
@davecharron New commits are in to hopefully resolve your raised issues. Let me know what you think. Thanks!

src/database/statement.ts
```diff
@@ -9,11 +9,18 @@ export default class Statement {
...
+   * @returns 
+   */
+  static delimName(name: string, fromUser = false) {
+    if (fromUser && name.startsWith(`"`) && name.endsWith(`"`)) return name;
+    if (fromUser === false && name.includes(` `)) return `"${name}"`;
     if (name.length <= 10) return name.toUpperCase();
-    else return `"${name}"`;
+    else if (fromUser) return `"${name}"`;
+    else return name;
   }
```
davecharron: I think this still needs tweaking.  This is not complete, but gives an idea of what I think is needed.
``` typescript
    if (fromUser) { // The name was input by the user
      // If already delimited, return it as-is
      if (name.startsWith(`"`) && name.endsWith(`"`)) return name;
      // If the value contains a space or decimal it needs to be delimited
      if (name.includes(` `) || name.includes(`.`)) return `"${name}"`;
      // Otherwise, fold to uppercase.  The user should have explicitly delimited if that was their intention.
      return name.toUpperCase();
    } else { // The name came from a catalog file query
      // If the name contains characters other than the valid variants, uppercase, digits, or underscores, it must be delimited
      // Note: looks like there is some good logic in vscode-ibmi\src\api\IBMi.ts\parserMemberPath that has a regex that is variant-character aware
      if (name.match(`[^A-Z0-9_@#$]`)) return `"${name}"`;
      // The name should already be uppercase coming from the catalog so just return it
      return name;
    }
```

src/database/schemas.ts
```diff
@@ -18,7 +33,7 @@ const typeMap = {
 export default class Database {
   static async getObjects(schema: string, type: SQLType, details: PageData = {}): Promise<BasicSQLObject[]> {
     // The database doesn't store the object name with the quotes.
-    schema = Statement.noQuotes(Statement.delimName(schema));
+    schema = Statement.noQuotes(Statement.delimName(schema, true));
```
davecharron, line 36: Calls to getObjects are likely driven by the user navigating the Schema Browser.  In that context, the schema name is not user input.  Further, it should already be properly delimited, so it only needs to be stripped of its quotes.
```suggestion
    schema = Statement.noQuotes(schema);
```

davecharron (changes requested):

src/database/schemas.ts
davecharron: Thinking more about this, I guess maybe the code is necessary for a non-delimited schema name, as it displays in lowercase and needs to go through `delimitName` to get upper-cased.  If so, disregard this suggestion.

src/database/schemas.ts
```diff
@@ -133,13 +148,10 @@ export default class Database {
...
 
   static async generateSQL(schema: string, object: string, type: SQLType): Promise<string> {
-    const content = instance.getContent();
-
     schema = Statement.noQuotes(Statement.delimName(schema));
     object = Statement.noQuotes(Statement.delimName(object));
 
-    // Remove plural and convert to uppercase. Needs work
-    let validType: string = type.substring(0, type.length-1).toUpperCase();
+    let validType: string = internalTypes[type];
```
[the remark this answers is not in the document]
worksofliam: This was totally my mistake. It was never `SQLType` in the first place (as we were passing in the 'internal type' like `TABLE` or `ALIAS` as inherited from a previous query). I am undoing my change and no longer using the `SQLType` type for the type parameter.

src/database/statement.ts
worksofliam: Your comments made it much easier for me to understand, thanks. I went ahead and used this as a base for the change and also implemented `validQsysName` which uses the variant characters from the connection to validate it. A new commit is in. Let me know what you think.

forstie:
Liam, 
This delimited name test isn't working. The table does not appear under tables.

image

forstie:
```
create schema "lower";
create table "lower"."lowt"("clo" integer);

```

forstie:
Thanks Liam for all the fixes. 
Here's some of what I tried:

`
create schema "CaMel";
create table "CaMel"."CoWt"("cLo" integer);
create view "CaMel"."CoWv" as select * from "CaMel"."CoWt";
create index "CaMel"."CoWi" on   "CaMel"."CoWt"("cLo");
create type "CaMel"."CoWt" as integer;
create procedure "CaMel"."CoWp" begin end;
create procedure "CaMel".proc1(p1 integer) begin end;
create procedure "CaMel"."CoWp2"("p1234" integer) begin end;
create procedure "CaMel".proc1(p1 integer) begin end;
create alias "CaMel"."Al's my pal today" for "CaMel"."CoWt";

create function "CaMel".fun1(p1 integer) returns integer return -1;
create function "CaMel"."fun2"("p1234" integer) returns integer return -1;
create function "CaMel"."Funny3"(p1 integer) returns integer return -1;
create variable  "CaMel"."Varbie was here today""!"" "   timestamp(12);
CREATE OR REPLACE VARIABLE "CaMel"."Varbie was here today""!""" FOR SYSTEM NAME "Varb0001" TIMESTAMP(12);`

worksofliam:
Thank you @forstie!

@davecharron Whenever you get the chance, can you perform a code review and some user testing? Thanks!

davecharron (approved):
All looks pretty good to me.  Ship it!

worksofliam:
@davecharron let’s go!!! Thank you 🙏

PR #42, opened 2023-07-13
bugfix: "Statement has already been fully run"

ThePrez:
Running the same statement multiple times resulted in an error saying "Statement has already been run" or "Statement has already been fully run."
Root cause is that we're reusing the web view to post result sets. We may want to revisit that core design decision in the future, but for now, we can simply reset the query object used in the view.


worksofliam (approved):
Tested with Jesse and no we are no longer seeing the message in the UI.

PR #43, opened 2023-07-13
Use EventEmitter to allow async server-side activity

ThePrez:


src/testing/jobs.ts
```diff
@@ -192,6 +192,37 @@ export const JobsSuite: TestSuite = {
...
+
+      await newJob.connect();
+      let stmt = `select * from QIWS.QCUSTCDT`;
+      let stmt2 = `values (job_name)`;
+      const resultAPromise = newJob.query(stmt).run();
+      const result2APromise = newJob.query(stmt2).run();
+      const resultBPromise = newJob.query(stmt).run();
+      const resultCPromise = newJob.query(stmt).run();
+      const result2BPromise = newJob.query(stmt2).run();
+      await Promise.all([resultAPromise,result2APromise, resultBPromise, resultCPromise, result2BPromise] ).then((values) => {
```
worksofliam, line 207: Can we do `const values = await Promise.all(...)` and use `try/catch/finally` with it.

PR #52, opened 2023-07-14
Allow column headers to be resizable

worksofliam:
Enables a horizontal drag on column headers.


forstie:
I tried to use this and could not find any way to resize my columns.

worksofliam:
Will come at a later time

PR #55, opened 2023-07-14
Use tokeniser to select value statement group

worksofliam:
SQL runner now supports:

* create+begin statements (procedures/functions mostly)
* compound statements!


forstie:
Tested and verified.
Great job!

PR #56, opened 2023-07-14
Status Bar item with commitment controls

worksofliam:
A new status bar item has been added that will:

* Have quick access to the job log and connection settings view
* Changes colour if there are pending uncommitted changes when using commitment control
* Shows how many many changes along with commit and rollback buttons.
* Stops jobs from ended when there are pending changes remaining without deciding what to do with them.

image

To test:

* Change isolation level from none
* Set auto commit to false

### DIFF
--- a/src/connection/sqlJob.ts
+++ b/src/connection/sqlJob.ts
@@ -11,6 +11,12 @@ export enum JobStatus {
   Active = "active",
   Ended = "ended"
 }
+
+export enum TransactionEndType {
+  COMMIT,
+  ROLLBACK
+}
+
 interface ReqRespFmt {
   id: string
 };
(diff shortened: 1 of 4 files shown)


forstie:
I tested this and had a couple of comments. Basically, the support works well.
1) The connection properties are a little complex in that autocommit can trip the user up.  I think this is planned to change.
2) Change  Revert to Rollback
3) Suggested a change to the pending or not pending text.
with xx(pending) as (
    select count(*) 
  from qsys2.db_transaction_info
  where JOB_NAME = qsys2.job_name and
        (local_record_changes_pending = 'YES' or
          local_object_changes_pending = 'YES')
) select case when pending = 1 then 'Pending Transaction'
else 'No pending transaction' end as Tranaction_monitor
from xx;

worksofliam:
Sent the updated VSIX with changes for 2 & 3.

forstie:
Changes tested and verified. Well done!

PR #60, opened 2023-07-18
Feature/save_config

worksofliam:
Adds the ability to right click on any job and save it's settings with a custom name to make it easier to create a job with matching settings again.

* Can right click on any job to save the settings into new configuration
   * Will be prompted for custom name
   * Current job will be renamed to the new name provided
* Saved configs appear in the SQL Job Manager view under a new tree item
* Clicking on the saved configuration will spin up a new job with the settings from the configuration
* Saved configurations have context options to edit and delete the config


worksofliam:
Sent @forstie an updated build with his suggestion: 

image

forstie:
Changes tested and confirmed.
Works great!
Thanks for the changes.

PR #66, opened 2023-07-20
Feature/cursor_autoclose

worksofliam:
This PR implements a cleanup process that is used so the user can swiftly run new statements with cursors and the backend correctly closes them as they run new ones.

This is implemented with a system that automatically closes flagged statements with `autoClose`, which should typically be used on user run statements. A test case has been written to prove that it does cleanup cursors.

We now automatically close the cursor when:

* Running statements through `JobManager.runSQL` (manually using `close`)
* Running paging queries in the results view (using the new `autoClose` method)

### DIFF
--- a/src/connection/query.ts
+++ b/src/connection/query.ts
@@ -17,7 +17,7 @@ export class Query<T> {
   private isCLCommand: boolean;
   private state: QueryState = QueryState.NOT_YET_RUN;
 
-  public shouldAutoClose: boolean | undefined;
+  public shouldAutoClose: boolean;
 
   constructor(private job: SQLJob, query: string, opts: QueryOptions = { isClCommand: false, parameters: undefined, autoClose: false }) {
     this.job = job;
--- a/src/connection/query.ts
+++ b/src/connection/query.ts
@@ -124,7 +124,6 @@ export class Query<T> {
   }
 
   public async close() {
-    // TODO: close the cursor
     this.state = QueryState.RUN_DONE;
 
     if (this.correlationId) {
--- a/src/connection/query.ts
+++ b/src/connection/query.ts
@@ -17,7 +17,7 @@ export class Query<T> {
   private isCLCommand: boolean;
   private state: QueryState = QueryState.NOT_YET_RUN;
 
-  public autoClose: boolean | undefined;
+  public shouldAutoClose: boolean | undefined;
 
   constructor(private job: SQLJob, query: string, opts: QueryOptions = { isClCommand: false, parameters: undefined, autoClose: false }) {
     this.job = job;
@@ -25,7 +25,7 @@ export class Query<T> {
     this.parameters = opts.parameters;
     this.sql = query;
     this.isCLCommand = opts.isClCommand;
-    this.autoClose = opts.autoClose;
+    this.shouldAutoClose = opts.autoClose;
 
     Query.globalQueryList.push(this);
   }
@@ -46,7 +46,7 @@ export class Query<T> {
     // First, let's check to see if we should also cleanup
     // any cursors that remain open, and we've been told to close
     for (const query of this.globalQueryList) {
-      if (query.autoClose) {
+      if (query.shouldAutoClose) {
         closePromises.push(query.close())
       }
     };
--- a/src/connection/query.ts
+++ b/src/connection/query.ts
@@ -124,9 +124,8 @@ export class Query<T> {
   }
 
   public async close() {
-    this.state = QueryState.RUN_DONE;
-
-    if (this.correlationId) {
+    if (this.correlationId && this.state !== QueryState.RUN_DONE) {
+      this.state = QueryState.RUN_DONE;
       let queryObject = {
         id: SQLJob.getNewUniqueRequestId(`sqlclose`),
         cont_id: this.correlationId,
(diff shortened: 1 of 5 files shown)


src/connection/query.ts
```diff
@@ -12,28 +12,57 @@ export class Query<T> {
...
+    for (const query of this.globalQueryList) {
+      if (query.autoClose) {
+        closePromises.push(query.close())
+      }
+    };
+
+    await Promise.all(closePromises);
+
+    // Automatically remove any queries done and dusted. They're useless.
+    this.globalQueryList = this.globalQueryList.filter(q => q.getState() !== QueryState.RUN_DONE);
```
ThePrez, line 57: Is there any flows here the UI might still be open for one of these queries?

src/views/results/index.ts
```diff
@@ -33,17 +33,17 @@ class ResultSetPanelProvider {
...
-        const content = instance.getContent();
-        const config = instance.getConfig();
         let data = [];
 
         let queryObject = Query.byId(message.queryId);
         try {
-          if (undefined === queryObject) {
-            let query = await JobManager.getPagingStatement(message.query, { isClCommand: message.isCL });
+          if (queryObject === undefined) {
+            Query.cleanup();
```
ThePrez, line 41: We will need to revisit this if we ever allow multiple result tabs like ACS does
(for now, a code comment indicating such is a good idea)

src/connection/query.ts
```diff
@@ -12,28 +12,57 @@ export class Query<T> {
...
-  constructor(private job: SQLJob, query: string, opts: QueryOptions = {isClCommand: false, parameters: undefined}) {
+  public autoClose: boolean | undefined;
+
+  constructor(private job: SQLJob, query: string, opts: QueryOptions = { isClCommand: false, parameters: undefined, autoClose: false }) {
     this.job = job;
     this.isPrepared = (undefined !== opts.parameters);
     this.parameters = opts.parameters;
     this.sql = query;
     this.isCLCommand = opts.isClCommand;
+    this.autoClose = opts.autoClose;
```
ThePrez, line 28: Personal preference and precedence in this part is that boolean member vars start with "is"... So `this.isAutoClose`

src/connection/query.ts
```diff
@@ -12,28 +12,57 @@ export class Query<T> {
...
   private sql: string;
   private isPrepared: boolean = false;
-  private parameters: any[];
+  private parameters: any[]|undefined;
   private rowsToFetch: number = 100;
   private isCLCommand: boolean;
   private state: QueryState = QueryState.NOT_YET_RUN;
 
-  constructor(private job: SQLJob, query: string, opts: QueryOptions = {isClCommand: false, parameters: undefined}) {
+  public autoClose: boolean | undefined;
```
ThePrez, line 20: An undefined boolean seems unconventional, and you always define it in the constructor, so I think the `| undefined` here can go away (?)

src/connection/query.ts
```diff
@@ -84,20 +114,33 @@ export class Query<T> {
...
       this.state = QueryState.ERROR;
       throw new Error(queryResult.error || `Failed to run query (unknown error)`);
     }
     return queryResult;
   }
+
   public async close() {
     this.state = QueryState.RUN_DONE;
+
+    if (this.correlationId) {
```
ThePrez, line 129: should be 
```typescript
if (this.correlationId && this.state !== QueryState.RUN_DONE)
```
and of course moved above the preceding assignment

ThePrez (changes requested):

src/connection/query.ts
worksofliam: Are you suggesting that `RUN_DONE` indicates that the cursor is closed?

src/connection/query.ts
worksofliam: I agree that this should be more explicit, but I don't know if `is` is right. I will try `shouldAutoClose`. Lmk what you think

src/connection/query.ts
worksofliam: Undefined boolean absolutely seems odd. But since `QueryOptions.autoClose` is `boolean?` (the same as `boolean|undefined`), I did it here. But you are right, both `false` and `undefined` are falsey values.

src/connection/query.ts
worksofliam: Not that I am aware off. Really if the value is `RUN_DONE` then we should be removing it from the array. `cleanup` will set some queries to `RUN_DONE` also.

src/connection/query.ts
ThePrez: `shouldAutoClose` is perfect

src/connection/query.ts
ThePrez: I agree (and this, I think, is tied to the future multi-tabs discussion)

src/connection/query.ts
ThePrez: yes. If it's `RUN_DONE` the cursor is closed

worksofliam: future PR is improved enum values :eyes:

ThePrez (approved):

src/connection/query.ts
ThePrez: resolved at https://github.com/halcyon-tech/vscode-db2i/pull/66/commits/96d3a4b5e79ac1c0c3bd5d082b2bc52406496a96

PR #74, opened 2023-07-21
re-check for server updates if not installed

ThePrez:


src/connection/manager.ts
```diff
@@ -117,6 +117,7 @@ export class SQLJobManager {
     if (ServerComponent.isInstalled() && selected) {
       return selected.job.query<T>(query, opts);
     } else if(!ServerComponent.isInstalled()) {
+      ServerComponent.checkForUpdate();
```
davecharron, line 120: Should the update status be checked here?  If successful, wouldn't we want to avoid throwing the error?

ThePrez: Yes. I was thinking the update happens asynchronously and it's reasonable for a user to "try again" after getting the error message. But we should response-check here and make it better

ThePrez: https://github.com/halcyon-tech/vscode-db2i/pull/74/commits/dd86d96b4939ffe4d593ec167f249966319bc0ba

ThePrez: marking resolved, reopen if needed

src/connection/manager.ts
```diff
@@ -112,15 +113,19 @@ export class SQLJobManager {
...
-      ServerComponent.checkForUpdate();
+      let updateResult = await ServerComponent.checkForUpdate();
+      if(UpdateStatus.JUST_UPDATED === updateResult) {
+        setTimeout(() => {JobManagerView.setVisible(true)});
+        return this.getPagingStatement(query, opts);
+      }
       throw new Error(`Database server component is required. Please see documentation for details.`);
     }else {
-      throw new Error(`Active SQL job is required. Please spin one up first.`);
+      throw new Error(`Active SQL job is required. Please spin one up in the 'SQL Job Manager' view and try again.`);
```
davecharron, line 128: Wondering why we couldn't spin up the job for them.  If they have multiple JDBC configs saved we wouldn't know which to choose, but it they had one or none, seems like we could maybe start the job.  Just a thought, no action required.

Code looks good, tested the flow and it seems fine.

davecharron (approved):

src/connection/manager.ts
ThePrez: I agree, but we should only do so if the autostart setting is set. change at https://github.com/halcyon-tech/vscode-db2i/pull/74/commits/dae5229fd8e159bfba19b83597e6ea72f2d6cd12

src/connection/manager.ts
```diff
@@ -117,10 +117,13 @@ export class SQLJobManager {
...
       return selected.job.query<T>(query, opts);
-    } else if(!ServerComponent.isInstalled()) {
+    } else if (!ServerComponent.isInstalled()) {
       let updateResult = await ServerComponent.checkForUpdate();
-      if(UpdateStatus.JUST_UPDATED === updateResult) {
-        setTimeout(() => {JobManagerView.setVisible(true)});
+      if (UpdateStatus.JUST_UPDATED === updateResult) {
+        JobManagerView.setVisible(true);
+        if (`new`=== Configuration.get<string>(`alwaysStartSQLJob`)) {
+          await commands.executeCommand(`vscode-db2i.jobManager.newJob`);
```
davecharron, line 125: When I pulled down the updates, my VS Code wasn't happy about these lines.
image

I needed to add imports to satisfy it.
``` typescript
import { commands } from "vscode";
import Configuration from "../configuration";
```

src/connection/manager.ts
```diff
@@ -117,10 +117,13 @@ export class SQLJobManager {
...
     if (ServerComponent.isInstalled() && selected) {
       return selected.job.query<T>(query, opts);
-    } else if(!ServerComponent.isInstalled()) {
+    } else if (!ServerComponent.isInstalled()) {
       let updateResult = await ServerComponent.checkForUpdate();
-      if(UpdateStatus.JUST_UPDATED === updateResult) {
-        setTimeout(() => {JobManagerView.setVisible(true)});
+      if (UpdateStatus.JUST_UPDATED === updateResult) {
+        JobManagerView.setVisible(true);
+        if (`new`=== Configuration.get<string>(`alwaysStartSQLJob`)) {
```
davecharron, line 124: For `alwaysStartSQLJob` to be set to `new` rather than `ask`, the user has to have installed the server component at some point prior.  Upon installation they would have been prompted to start the server job, and would have had to answer Always.

Given that, consider my scenario.

At startup, I got the prompt to install the server component, but I cancelled it, :smirk: then proceeded to try to run some SQL, and got the prompt to install again.  This time the install was performed, but the server job wasn't started automatically because the setting is still set to `ask`.  I still got the `Active SQL job is required. Please spin one up in the 'SQL Job Manager' view and try again.` message.

Could it have presented the prompt rather than the error message?
image

davecharron (changes requested):

src/connection/manager.ts
ThePrez: sorry
https://github.com/halcyon-tech/vscode-db2i/pull/74/commits/685feb475eee0eaf9169f9bf555c64df3d6d5f29

src/connection/manager.ts
ThePrez: resolved at https://github.com/halcyon-tech/vscode-db2i/pull/74/commits/090952b7e35ae5fd0b607851ad95b87984251ba5

worksofliam (approved):
@davecharron I went ahead and also added a small commit to make the UX of spinning up a job nicer when none are active.

I'd appreciate it if you could give this another look over, as well as another test. Thanks!

davecharron (approved):
Seems to be in good shape.

PR #85, opened 2023-07-24
fixup failing testcases on main

ThePrez:


worksofliam (approved):
I added missing commits from `main` and also removed a test that compared the new server component to the legacy `runSQL` API since we do not use that anymore.

All tests are passing. Thank you!

PR #86, opened 2023-07-25
Use server-side CLI and set client special registers

ThePrez:


worksofliam (changes requested):
I see what you are going for here @ThePrez - putting the version into the runtime VSIX. I believe we can (and should) do this without adding another dev dep.

In `webpack.config.js`, you can `const package = require('./package.json')` and then use the `new webpack.DefinePlugin` plugin to register an environment variable (perhaps `DB2I_VERSION) which could hold the value of `package.version`. We do this already with `process.env.DEV`.

Let me know what you think or if you have any questions.

ThePrez:
Thanks for the guidance!

ThePrez:
fixes #48

worksofliam (changes requested):
Is it possible to write a test to ensure the client register is set? I am sure there's a way to get them with an SQL statement.

PR #90, opened 2023-07-27
Build result view columns from query metadata

ThePrez:


src/views/results/index.ts
```diff
@@ -50,6 +50,7 @@ class ResultSetPanelProvider {
           this._view.webview.postMessage({
             command: `rows`,
             rows: queryResults.data,
+            columnList: queryResults.metadata.columns.map(x=>x.name),
```
worksofliam, line 53: This is likely an issue because `metadata` is not (and should not) be returned on every fetch.

worksofliam (changes requested):
@ThePrez Little issue. Works perfect for the first fetch when the meta data from the server is returned, but when trying to fetch the next page it fails with this error:

image

This is with a simple `select * from qsys2.columns`.

src/views/results/index.ts
worksofliam: edit: looks like you solved this exact issue in PR #93

ThePrez: Plucked to https://github.com/halcyon-tech/vscode-db2i/pull/90/commits/c5178ef0a0b719f6dbc7280a2386adadeb1ce21f

worksofliam (approved):

PR #93, opened 2023-07-27
Handle terse data format from server for SQL results

### LINKED ISSUE #92: Handle short form data fetching from server
Currently, SQL query results are fetched from the server as an array of objects (each object has an attribute corresponding to the column name). 

Instead, to save transport overhead, just use a two-dimensional array of data. This saves the JSON payload from repeating the same column names over and over.

ThePrez:
fixes #92 
wait for #90

### DIFF
--- a/src/connection/manager.ts
+++ b/src/connection/manager.ts
@@ -89,7 +89,13 @@ export class SQLJobManager {
 
     return (this.selectedJob >= 0);
   }
-
+  /**
+   * Runs SQL
+   * @param query the SQL query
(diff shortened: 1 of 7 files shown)


worksofliam:
@ThePrez Any chance you can fix up these merge conflicts before I continue with my testing?

worksofliam (approved):
Fixed a signature change that happened due to a merge. All the tests are passing and thank you for adding this mode. It will prove useful for larger result sets I am sure.

image

PR #94, opened 2023-07-27
Overloaded routine names

### LINKED ISSUE #36: Schema browser need better support for procedure overloading

davecharron:
Add support for overloaded routine names, fixes #36

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -1,3 +1,5 @@
+name: Build and publish a release
+
 on:
   release:
     types: [created]
--- a/.github/workflows/webpack_ci.yaml
+++ b/.github/workflows/webpack_ci.yaml
@@ -29,11 +29,24 @@ jobs:
         npm install -g vsce
         vsce package
 
-    - name: Copy dist artifacts to staging directory
-      run: mkdir staging && cp ./*.vsix staging
+    - name: get .vsix name
+      run: |
+        echo "vsix_name=$(echo *.vsix)" >> $GITHUB_ENV
+        cat $GITHUB_ENV
 
     - name: Upload dist artifacts
       uses: actions/upload-artifact@v3
       with:
-        name: Package
-        path: staging
+        name: test_vsix
+        path: ${{env.vsix_name}}
+          
+    - name: Post comment
+      uses: actions/github-script@v5
+      with:
+        script: |
+          github.rest.issues.createComment({
+            issue_number: context.issue.number,
+            owner: context.repo.owner,
+            repo: context.repo.repo,
+            body: '👋  A new build is available for this PR based on ${{ github.event.pull_request.head.sha }}.\n * [Download here.](https://github.com/halcyon-tech/vscode-db2i/actions/runs/${{ github.run_id }})\n'
+          })
--- a/global.d.ts
+++ b/global.d.ts
@@ -39,6 +39,7 @@ interface BasicSQLObject {
   type: string;
   schema: string;
   name: string;
+  specificName: string;
   text: string;
   system: {
     schema: string;
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "vscode-db2i",
   "displayName": "Db2 for IBM i",
   "description": "Db2 for IBM i tools in VS Code",
-  "version": "0.3.2",
+  "version": "0.3.3",
   "preview": true,
   "engines": {
     "vscode": "^1.70.0"
--- a/src/connection/manager.ts
+++ b/src/connection/manager.ts
@@ -90,12 +90,12 @@ export class SQLJobManager {
     return (this.selectedJob >= 0);
   }
 
-  async runSQL<T>(query: string, parameters: any[] = []): Promise<T[]> {
+  async runSQL<T>(query: string, opts: QueryOptions = {}): Promise<T[]> {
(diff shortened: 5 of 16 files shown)


davecharron:
#### Functional changes in the Schema Browser
- Clicking a routine (function or procedure) properly expands to display the correct parameters for the selected routine
image
- Tool tips on routines include the specific name for differentiation when names are overloaded
  Setting the tool tip rather than changing the tree item label keeps the view cleaner yet still provides the extra info
image
- Generate SQL action changed to use the specific name so the results are only generated for the selected routine
  ( previously all routines with the same name were generated )
- Delete action changed to use the specific name so the selected routine can be deleted
  ( previously delete failed when multiple routines had the same name )

#### Setup script for testing
```sql
CREATE SCHEMA OVERLOAD;
CREATE OR REPLACE FUNCTION OVERLOAD.MULTI_FUNC (
            P1 INTEGER DEFAULT 13
    )
    RETURNS INTEGER
    LANGUAGE SQL
    MODIFIES SQL DATA
    CONCURRENT ACCESS RESOLUTION DEFAULT
    FENCED
    NOT DETERMINISTIC
    CALLED ON NULL INPUT
    EXTERNAL ACTION NOT
    SECURED
    SPECIFIC OVERLOAD.MULTI_SQ
    RETURN P1 * P1;

LABEL ON ROUTINE OVERLOAD.MULTI_FUNC(INT) IS 'Accepts One Integer and Returns Its Square';

COMMENT ON PARAMETER ROUTINE OVERLOAD.MULTI_FUNC (INT) (P1 IS 'Value to Square');


CREATE OR REPLACE FUNCTION OVERLOAD.MULTI_FUNC (
            P1 INTEGER DEFAULT 13,
            P2 INTEGER DEFAULT 74
        )
    RETURNS INTEGER
    LANGUAGE SQL
    MODIFIES SQL DATA
    CONCURRENT ACCESS RESOLUTION DEFAULT
    FENCED
    NOT DETERMINISTIC
    CALLED ON NULL INPUT
    EXTERNAL ACTION NOT
    SECURED
    SPECIFIC OVERLOAD.MULTI_PROD
    RETURN P1 * P2;

LABEL ON ROUTINE OVERLOAD.MULTI_FUNC(INT, INT) IS 'Accepts Two Integer and Returns the Product';

COMMENT ON PARAMETER ROUTINE OVERLOAD.MULTI_FUNC (INT, INT) (P1 IS 'First Value', P2 IS 'Second Value');


/* Creating OVERLOAD.MULTI_PROC [Procedure] */
CREATE OR REPLACE PROCEDURE OVERLOAD.MULTI_PROC ()
        LANGUAGE SQL
        MODIFIES SQL DATA
        SPECIFIC OVERLOAD.MULTI_MONS
        PROGRAM TYPE SUB
        CONCURRENT ACCESS RESOLUTION DEFAULT
        DYNAMIC RESULT SETS 1
        OLD SAVEPOINT LEVEL
        COMMIT ON RETURN NO
BEGIN
    DECLARE MONITORS CURSOR WITH RETURN FOR
        SELECT *
            FROM QUSRSYS.QAUGDBPMD2;
    OPEN MONITORS;
END;

/* Setting label text for OVERLOAD.MULTI_PROC */
LABEL ON SPECIFIC ROUTINE OVERLOAD.MULTI_MONS IS 'Returns the Entire List of Monitors';

/* Creating OVERLOAD.MULTI_PROC [Procedure] */
CREATE OR REPLACE PROCEDURE OVERLOAD.MULTI_PROC (
                IN CREATOR CHARACTER(10) DEFAULT ''
            )
        LANGUAGE SQL
        MODIFIES SQL DATA
        SPECIFIC OVERLOAD.MULTI_MYMONS
        PROGRAM TYPE SUB
        CONCURRENT ACCESS RESOLUTION DEFAULT
        DYNAMIC RESULT SETS 1
        OLD SAVEPOINT LEVEL
        COMMIT ON RETURN NO
BEGIN
    DECLARE MONITORS CURSOR WITH RETURN FOR
        SELECT *
            FROM QUSRSYS.QAUGDBPMD2
            WHERE "Created by" = CREATOR;
    OPEN MONITORS;
END;

/* Setting label text for OVERLOAD.MULTI_PROC */
LABEL ON SPECIFIC ROUTINE OVERLOAD.MULTI_MYMONS IS 'Returns the List of My Monitors';

/* Setting comment text for OVERLOAD.MULTI_PROC */
COMMENT ON PARAMETER SPECIFIC ROUTINE OVERLOAD.MULTI_MYMONS (CREATOR IS 'User That Created the Monitor');
```

worksofliam (changes requested):
The code is fab. Let's get that instance changes where parameter markers could be used and then I will have another round of testing for you.

davecharron:
I think I've made the required changes.  Please take another look.

worksofliam (changes requested):
There is one part I'd love to see in this PR which is some tests cases that cover:

* creating the environment with the example you provided
* fetching those new functions and procedures
* checking that you can differentiate between the two functions/procedures in the API

I am happy to write those and you are welcome to also. See this test suite for how we tested the delimiter PR: 
https://github.com/halcyon-tech/vscode-db2i/blob/main/src/testing/database.ts

worksofliam (approved):
Tests are passing and also works in the UI! 

Great work @davecharron!

image

PR #96, opened 2023-07-31
Add test for round-tripping extended characters

ThePrez:
it fails lol

### DIFF
--- a/src/testing/jobs.ts
+++ b/src/testing/jobs.ts
@@ -100,6 +100,22 @@ export const JobsSuite: TestSuite = {
       newJob.close();
     }},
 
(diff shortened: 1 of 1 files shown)


ThePrez:
ok now it works

worksofliam (changes requested):
@ThePrez Code looks ok, but I'd love to see a better PR description which will provide better context for the PR. Thanks

ThePrez:
Sorry, @worksofliam , the initial PR was only to add a testcase for the extended characters case but I decided to pump the fix in as well. Original PR description updated to better reflect the nature of the change

PR #105, opened 2023-08-04
Validate user-entered schema / Browse to add schemas

### LINKED ISSUE #61: Validate schema when adding to schema browser
Either:

1. validate the user input to ensure the schema exists, or
2. show a list of all schemas the user can filter through instead

davecharron:
Ready for review and some testing.  Fixes #61 

The input prompt for adding a schema to the Schema Browser will now:
1. Validate the input to prevent duplicate entries.  Validation is performed on every keystroke, so checking for duplicates can be done in the validator, but verifying existence needs to wait until the user presses Enter.
image
2. Verify existence of the schema entered, displaying an error when it does not.  Unfortunately the `InputBox` construct does not seem to allow for validation upon the user pressing Enter, so for now it simply displays an error.  We could swizzle it to re-display the `InputBox` with the value prefilled and additional validation code that would highlight the value as an error.
image

Also added a schema browser for quickly adding one or more schemas.
image

Display includes the system schema name if different from the SQL name, as well as the text description.
image

List filtering capability is provided by the `vscode.window.showQuickPick` API
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -163,6 +163,12 @@
         "icon": "$(add)"
       },
       {
+        "command": "vscode-db2i.browseForSchemasToAddToSchemaBrowser",
+        "title": "Browse for Schemas to Add to Schema Browser",
+        "category": "Db2 for i",
+        "icon": "$(list-selection)"
+      },
+      {
         "command": "vscode-db2i.removeSchemaFromSchemaBrowser",
         "title": "Remove schema from view",
         "category": "Db2 for i",
@@ -389,6 +395,11 @@
       "view/title": [
         {
           "command": "vscode-db2i.addSchemaToSchemaBrowser",
+          "group": "navigation",
+          "when": "view == schemaBrowser"
+        },
+        {
+          "command": "vscode-db2i.browseForSchemasToAddToSchemaBrowser",
           "group": "navigation",
           "when": "view == schemaBrowser"
         },
--- a/package.json
+++ b/package.json
@@ -157,14 +157,8 @@
         "icon": "$(refresh)"
       },
       {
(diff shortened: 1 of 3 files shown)


forstie:
Note... this one cannot be fully verified because there's no support for Database IASPs.

forstie:
Confirmed!
Some comments were shared in slack.

davecharron:
Made changes to eliminate the Add button.  Only the manage button is now displayed.
image

The chooser indicates the current list, and those available.
image

Upon clicking `OK` the Schema Browser will update to show only those selected on the chooser.

Schema system name only displayed if different from the SQL name.  Moved from the `description` to the `details` section of the list item, which renders on a separate line.  Filtering works against both the name and system name.
image

worksofliam (changes requested):
Beautiful work. This is a serious improvement over what we had before and I love the use of `createQuickPick`. Two issues I have found that need some work:

1. End all SQL jobs, then click the new button to select the schemas from the list. When prompted if you want to create a new job, **say no**, which results in an error from your new command. This should likely be handled inside of your new command with a nicer error message.
2. I was on a slow connection this morning, and clicking on the new button didn't do anything, which made me think it's probably fetching the list. The user might not think that, so I think some progress indicator would be useful to show that it is actually doing something. Check out the `window.withProgress` API for that.

Thanks and lmk if you have any questions.

davecharron:
Changes made to address the issues.

worksofliam (approved):
Thanks for clearing up my issues - it works as expected now. The loading indicator is well appreciated because the system I am using is seriously slow. Also, gives me a nicer error if no job is available.

image

PR #108, opened 2023-08-13
Add update count to user message when no results are returned

### LINKED ISSUE #101: Q : Change "Query executed with no result set returned." to actually return how many rows were affected
Hi
I have a question, is there any way to enable in the extension so it actually returns how many rows were affected i.e. using sql UPDATE in the ibm i results window. I was looking all over and could not find it in the extension settings. 

Thank you
Matt

ThePrez:
Fixes #101


worksofliam (approved):
Working great! Lots of feedback for such a small change.

image

@ThePrez For another PR perhaps, but do we need a test to check to `update_count`?

PR #112, opened 2023-08-28
Feature/improved_index_advice

worksofliam:
Scott made a suggestion that we should use condenced index advice instead and even provided an improved statement for us to use.


forstie:
After another reload of the vsix file, the support worked great.
I did notice a problem with the data, which Liam has pursued with Jesse.
Great going!

PR #128, opened 2023-09-21
Fix/dumb asp support

worksofliam:
* Use `tcp` mode if a database name is selected
* Show dropdown of available databases/iASP

image


forstie:
Works like a champ!
Thank ya.

forstie (approved):

PR #135, opened 2023-09-22
`service_info` available from Examples view

worksofliam:
* Adds 'IBM i (SQL) Services' to Examples view. This is populared from the `service_info` object.


davecharron (approved):
Looks good.

PR #140, opened 2023-09-23
Feature/multiple_object_types

worksofliam:
* Adds ability to filter objects in a schema
* getObjects can now fetch multiple types at once
* Fixe bug where procedure/function parameters were not returned due to specific name check

image

### DIFF
--- a/src/database/schemas.ts
+++ b/src/database/schemas.ts
@@ -13,122 +13,113 @@ const typeMap = {
   'aliases': [`A`]
 };
 
-export const AllSQLTypes: SQLType[] = ["schemas", "tables", "views", "aliases", "constraints", "functions", "variables", "indexes", "procedures", "sequences", "packages", "triggers", "types"];
-
 export default class Database {
   /**
    * @param schema Not user input
    */
-  static async getObjects(schema: string, types: SQLType[], details: PageData = {}): Promise<BasicSQLObject[]> {
-    let selects: string[] = [];
+  static async getObjects(schema: string, type: SQLType, details: PageData = {}): Promise<BasicSQLObject[]> {
     let query : string;
+    switch (type) {
+      case `schemas`:
+        query = [
+          `select SCHEMA_NAME as NAME, SCHEMA_TEXT as TEXT, SYSTEM_SCHEMA_NAME as SYS_NAME `,
+          `from QSYS2.SYSSCHEMAS`,
+          details.filter ? `where SCHEMA_NAME = '${details.filter}' or SYSTEM_SCHEMA_NAME = '${details.filter}'` : ``,
+          `order by QSYS2.DELIMIT_NAME(SCHEMA_NAME) asc`
+        ].join(` `);
+          break;
+    case `tables`:
+    case `views`:
+    case `aliases`:
+      query = [
+        `select TABLE_NAME as NAME, TABLE_TEXT as TEXT, BASE_TABLE_SCHEMA as BASE_SCHEMA, BASE_TABLE_NAME as BASE_OBJ, SYSTEM_TABLE_NAME as SYS_NAME, SYSTEM_TABLE_SCHEMA as SYS_SCHEMA `,
+        `from QSYS2.SYSTABLES`,
+        `where TABLE_SCHEMA = '${schema}' and TABLE_TYPE in (${typeMap[type].map(item => `'${item}'`).join(`, `)}) ${details.filter ? `and TABLE_NAME like '%${details.filter}%'`: ``}`,
+        `order by TABLE_NAME asc`
+      ].join(` `);
+      break;
 
-    for (const type of types) {
-      switch (type) {
-        case `schemas`:
-          selects.push([
-            `select '${type}' as OBJ_TYPE, SCHEMA_NAME as NAME, SCHEMA_TEXT as TEXT, SYSTEM_SCHEMA_NAME as SYS_NAME `,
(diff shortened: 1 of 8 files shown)


davecharron:
I'm not seeing the schema filter value displaying next to the schema name.

davecharron:
Schema Browser management no longer works.
image

src/database/schemas.ts
```diff
@@ -13,132 +13,152 @@ const typeMap = {
...
-        `where USER_DEFINED_TYPE_SCHEMA = '${schema}' ${details.filter ? `and USER_DEFINED_TYPE_NAME like '%${details.filter}%'`: ``}`,
-        `order by USER_DEFINED_TYPE_NAME asc`
-      ].join(` `);
-      break;
+
+    for (const type of types) {
+      switch (type) {
+        case `schemas`:
+          selects.push([
+            `select '${type}' as OBJ_TYPE, SCHEMA_NAME as NAME, SCHEMA_TEXT as TEXT, SYSTEM_SCHEMA_NAME as SYS_NAME, SYSTEM_TABLE_SCHEMA as SYS_SCHEMA, `,
```
davecharron, line 30: `SYSTEM_TABLE_SCHEMA` is not a column in `SYSSCHEMAS`

src/views/schemaBrowser/schemaBrowser.ts
```diff
@@ -431,10 +443,22 @@ export default class schemaBrowser {
...
 
       if (element instanceof Schema) {
-        items = getSchemaItems(element.schema);
+
+        let filterValue = this.filters[element.schema];
+        if (filterValue) {
+          filterValue = Statement.noQuotes(Statement.delimName(filterValue, true));
+          const filteredObjects = await Schemas.getObjects(element.schema, AllSQLTypes, {filter: filterValue});
+          items = filteredObjects.map(obj => new SQLObject(obj));
+
```
davecharron, line 452: Need to set the schema node description to reflect the current filter.

davecharron:
Parameters are not retrieved for overloaded routines when filter is applied.
No filter:
image
Filtered:
image

davecharron:
I wonder if we could do something to add more pop to the filtered schema node with additional text or color/highlighting.

Should the word 'filter' be included somewhere, instead of just showing the filter string within parenthesis?
Ex: `OVERLOAD Filter: multi`

Text can be highlighted via the TreeItemLabel interface.
Ex: `super(currentFilterLabel ? {label:name, highlights:[[0,name.length]]} : name, vscode.TreeItemCollapsibleState.Collapsed);`
But the highlighting it does would be better for highlighting the search term within labels.
image

It'd be cool to colorize like the explorer does with changed source parts:
image

Check this out: https://stackoverflow.com/a/74493474/6890772

davecharron:
> > Parameters are not retrieved for overloaded routines when filter is applied. No filter: image Filtered: image
> 
> I realized this morning that the filter is likely being applied when the parameters are retrieved. I would suggest that the filter only be applied at the top object level.

Debugging shows that when the list is filtered the specific name for functions and procedures is not retrieved.

src/database/schemas.ts
```diff
@@ -13,132 +13,152 @@ const typeMap = {
...
+          `  '' as SPECNAME`,
+          `from QSYS2.SYSCST`,
+          `where CONSTRAINT_SCHEMA = '${schema}' ${details.filter ? `and CONSTRAINT_NAME like '%${details.filter}%'`: ``}`,
+          `order by CONSTRAINT_NAME asc`
+        ].join(` `));
+        break;
+  
+      case `functions`:
+        selects.push([
+          `select '${type}' as OBJ_TYPE, ROUTINE_NAME as NAME, SPECIFIC_NAME as SPECNAME, coalesce(ROUTINE_TEXT, LONG_COMMENT) as TEXT, `,
```
davecharron, line 62: Moving `, SPECIFIC_NAME as SPECNAME` to the last column position fixes the issue with specific name missing when retrieving parameters.
Same comment for the procedure query below.
General comment/rule is that with the `UNION` of all these queries, the column order within each should be the same.

src/views/schemaBrowser/schemaBrowser.ts
```diff
@@ -431,10 +445,22 @@ export default class schemaBrowser {
       const contextValue = element.contextValue;
 
       if (element instanceof Schema) {
-        items = getSchemaItems(element.schema);
+
+        let filterValue = this.filters[element.schema];
+        if (filterValue) {
+          filterValue = Statement.noQuotes(Statement.delimName(filterValue, true));
+          const filteredObjects = await Schemas.getObjects(element.schema, AllSQLTypes, {filter: filterValue});
```
davecharron, line 452: For a schema with a delimited name, `element.schema` includes the delimiting quotes.  Those need to be stripped off when passed into `getObjects`.

worksofliam: @davecharron that is what `noQuotes` does, unless you are referring to elsewhere in the code?

davecharron: On line 451 it's applying it to the filter value, not the schema name.

worksofliam: Fixed in latest commit. Took me a while to catch on.

davecharron (approved):
Not finding anything more to complain about.  :beers: :checkered_flag:

PR #142, opened 2023-09-25
Feature: Visual Explain

worksofliam:
* Adds new API to run VE on the server
* New API to convert flat VE data to nodes


forstie:
In my opinion, it should be easier to differentiate the name of a topic and the value for a topic.
image

forstie:
I cannot find index or statistics advice. This is table stakes for VE.

        {
          "type": 111,
          "title": "Index advised information",
          "value": ""
        },

forstie:
Seems that some users will expect keyboard shortcuts. 
Here's what Run SQL Scripts uses:
image

forstie:
When I use 'Explain without Running", the blue rocket starts taking off and never lands.
Since its not Running the query, the rocket shouldn't take off

image

forstie:
The use of this color is very difficult to read.
image

forstie:
Why does "ACS" appear ??
image

davecharron:
> In my opinion, it should be easier to differentiate the name of a topic and the value for a topic. image

I don't disagree with your observation, but the rendering of tree items is managed by VS Code.
https://code.visualstudio.com/api/references/vscode-api#TreeItem
> TreeItem
description?: string | boolean
**A human-readable string which is rendered less prominent.** When true, it is derived from [resourceUri](https://code.visualstudio.com/api/references/vscode-api#TreeItem.resourceUri) and when falsy, it is not shown.

davecharron:
> Why does "ACS" appear ?? image

It's part of the data returned by the server-side procedure that processes the monitor output.

davecharron:
> The use of this color is very difficult to read. image

Looks better in dark mode.  Will need to adjust colors for light mode.

davecharron:
> I cannot find index or statistics advice. This is table stakes for VE.
> 
> ```
>     {
>       "type": 111,
>       "title": "Index advised information",
>       "value": ""
>     },
> ```

The advised index information is shown in the node details, but we have not provided an aggregate view for the entire plan.  I can look at generating that report.  I am thinking it would open in an editor tab similar to exporting the VE data.

src/connection/sqlJob.ts
```diff
@@ -209,6 +221,27 @@ export class SQLJob {
...
+      sql: statement,
+      run: type === ExplainType.Run
+    }
+
+    const result = await this.send(JSON.stringify(explainRequest));
+
+    const explainResult: ExplainResults<any> = JSON.parse(result);
+
+    if (explainResult.success !== true) {
+      this.dispose();
```
ThePrez, line 237: This line and the next should be removed, as it unnecessarily closes the channel

PR #144, opened 2023-09-27
Fix/current_schema

worksofliam:
Makes set current schema option functional and will update the server job.

`SET SCHEMA` and it's variations only work with SQL naming.

### DIFF
--- a/src/language/providers/completionProvider.ts
+++ b/src/language/providers/completionProvider.ts
@@ -14,6 +14,7 @@ import CompletionItemCache, { changedCache } from "./completionItemCache";
 import Callable from "../../database/callable";
 import { ServerComponent } from "../../connection/serverComponent";
 import { env } from "process";
+import { getInstance } from "../../base";
 
 const completionItemCache = new CompletionItemCache();
 
@@ -524,7 +525,7 @@ export const completionProvider = languages.registerCompletionItemProvider(
   `sql`,
   {
     async provideCompletionItems(document, position, token, context) {
-      if (ServerComponent.isInstalled() && isEnabled()) {
+      if (ServerComponent.isInstalled() && isEnabled() && JobManager.getSelection()) {
         const trigger = context.triggerCharacter;
         const content = document.getText();
         const offset = document.offsetAt(position);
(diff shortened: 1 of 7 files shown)


src/views/results/index.ts
```diff
@@ -254,9 +249,29 @@ export function initialise(context: vscode.ExtensionContext) {
   )
 }
 
+export function statementMiddleware(statement: StatementInfo) {
+  switch (statement.type) {
+    case StatementType.Create:
+    case StatementType.Alter:
+      const ref = statement.refs[0];
+      const databaseObj =
+        statement.type === StatementType.Create && ref.createType.toUpperCase() === `schema`
```
davecharron, line 258: This will never be true, right?
`ref.createType.toUpperCase() === 'schema'`

src/language/providers/completionProvider.ts
```diff
@@ -543,6 +544,12 @@ export const completionProvider = languages.registerCompletionItemProvider(
...
 
 const getDefaultSchema = (): string => {
+  const instance = getInstance();
+  const config = instance.getConfig();
+
   const currentJob = JobManager.getSelection();
-  return currentJob && currentJob.job.options.libraries[0] ? currentJob.job.options.libraries[0] : `QGPL`;
+  const currentSchema = currentJob ? currentJob.job.options.libraries[0] : undefined;
+  
+  const schema = currentJob && currentSchema !== `*LIBL` && currentSchema !== undefined ? currentJob.job.options.libraries[0] : (config.currentLibrary || `QGPL`);
```
davecharron, line 553: Nitpicking here, eliminate the constant and just return the value.

davecharron (approved):
Changes seem reasonable

src/views/results/index.ts
worksofliam: Good spot. I will fix this

src/language/providers/completionProvider.ts
worksofliam: Good point. I think I put it there to debug.

worksofliam:
This PR is going to be re-written after a cleanup PR has been finished. Closing.

PR #152, opened 2023-10-02
APIs for embedded SQL

worksofliam:
* Statement API to get embedded areas
* Document API to cleanup embedded statements
* Tests for different embedded scenarios

PR #153, opened 2023-10-02
Host parameters treated like snippets

worksofliam:
When a statement is ran, but has host parameters, user will be expected to fix them up before the statement can be executed.

The statement will not execute and the user will be able to tab through the parameters to set the values.


worksofliam:
Depends on #152

worksofliam:
Nicely tested by @forstie and worked in a fix for a bug he found.

PR #156, opened 2023-10-16
Feature/self codes

ajshedivy:
Test Script:

```sql
-- SETUP, run once
CREATE SCHEMA SELFTEST
CREATE OR REPLACE TABLE SELFTEST.MYTBL (C1 INT, C2 VARCHAR(100), C3 TIMESTAMP, C4 DATE)
CREATE OR REPLACE TABLE SELFTEST.MYTBL2 (C1 INT, C2 VARCHAR(100), C3 TIMESTAMP, C4 DATE)
INSERT INTO SELFTEST.MYTBL VALUES (0, 'ADAM', CURRENT TIMESTAMP, CURRENT DATE)
INSERT INTO SELFTEST.MYTBL VALUES (1, 'LIAM', CURRENT TIMESTAMP + 1 SECOND, CURRENT DATE + 1 DAY)
INSERT INTO SELFTEST.MYTBL VALUES (2, 'RYAN', CURRENT TIMESTAMP + 2 SECOND, CURRENT DATE + 2 DAY)
INSERT INTO SELFTEST.MYTBL VALUES (3, NULL, CURRENT TIMESTAMP + 2 SECOND, CURRENT DATE + 2 DAY)

INSERT INTO SELFTEST.MYTBL2 VALUES (0, 'TIM', CURRENT TIMESTAMP, CURRENT DATE)
INSERT INTO SELFTEST.MYTBL2 VALUES (1, 'SCOTT', CURRENT TIMESTAMP + 1 SECOND, CURRENT DATE + 1 DAY)
INSERT INTO SELFTEST.MYTBL2 VALUES (2, 'JESSIE', CURRENT TIMESTAMP + 2 SECOND, CURRENT DATE + 2 DAY)
STOP

SET SCHEMA SELFTEST

-- SQL0138, get left-most 1001 characters of C2. C2 is only 100 characters long, so SQL0138 is produced
SELECT LEFT(C2, 1001) FROM SELFTEST.MYTBL

-- SQL0180, invalid format for date given
VALUES DATE('120-1231-12312')

-- SQL0181, valid format for date given, but the date doesn't exist
VALUES DATE('2023-45-45')

-- SQL0182, valid datetime exists, but context in which the timestamp is used is invalid (dividing int by timestamp doesn't make sense)
SELECT C3/C4 FROM SELFTEST.MYTBL

-- SQL0199, invalid syntax, missing predicate after WHERE and before ORDER BY
SELECT * FROM SELFTEST.MYTBL WHERE ORDER BY C1

-- SQL0203, C2 doesn't have a table name associated with it (T1 or T2), the database doesn't know what C2 to select
SELECT SELFTEST.MYTBL.C2 FROM SELFTEST.MYTBL T1
JOIN SELFTEST.MYTBL2 T2 ON
T1.C1 = T2.C1

-- SQL0204, trying to select from a table that doesn't exist
SELECT * FROM NOEXIST

-- SQL0206, column doesn't exist (but the table does)
SELECT NOCOL FROM SELFTEST.MYTBL

-- SQL0420, can't cast varchar to int
SELECT CAST(SELFTEST.MYTBL.C2 AS INT) FROM SELFTEST.MYTBL

-- SQL0551, user profile NOGOOD has no authority to access database file
CL: CRTUSRPRF USRPRF(NOGOOD)
SET SESSION AUTHORIZATION NOGOOD -- Swap to user profile with no permissions
SELECT * FROM SELFTEST.MYTBL2

DISCONNECT CURRENT -- Can't swap back to your normal profile once you've given up permissions, need to disconnect and reconnect
SET SCHEMA SELFTEST
CL: DLTUSRPRF USRPRF(NOGOOD) -- Clean up

-- SQL0811, subquery in update returns multiple rows, database doesn't know what row to use for update. Subquery must be correlated
UPDATE SELFTEST.MYTBL SET C1 = (SELECT C1 FROM SELFTEST.MYTBL2)




```

### DIFF
--- a/package.json
+++ b/package.json
@@ -318,6 +318,11 @@
         "category": "Db2 for i"
       },
       {
+        "command": "vscode-db2i.jobManager.getSelfErrors",
+        "title": "Get SELFCODES Errors",
+        "category": "Db2 for i"
+      },
+      {
         "command": "vscode-db2i.jobManager.newConfig",
         "title": "Save settings to config",
         "category": "Db2 for i",
@@ -368,6 +373,10 @@
         },
         {
           "command": "vscode-db2i.jobManager.enableTracing",
+          "when": "never"
+        },
+        {
+          "command": "vscode-db2i.jobManager.getSelfErrors",
           "when": "never"
         },
         {
@@ -553,6 +562,11 @@
           "command": "vscode-db2i.jobManager.getTrace",
           "when": "view == jobManager && viewItem == sqlJob",
           "group": "trace@2"
+        },
+        {
+          "command": "vscode-db2i.jobManager.getSelfErrors",
+          "when": "view == jobManager && viewItem == sqlJob",
+          "group": "trace@3"
         },
         {
           "command": "vscode-db2i.jobManager.newConfig",
--- a/package.json
+++ b/package.json
@@ -290,6 +290,12 @@
         "icon": "$(edit)"
       },
       {
+        "command": "vscode-db2i.jobManager.editSelfCodes",
+        "title": "Edit SELFCODES",
+        "category": "Db2 for i",
+        "icon": "$(bracket-error)"
+      },
+      {
         "command": "vscode-db2i.jobManager.copyJobId",
         "title": "Copy Job Name",
         "category": "Db2 for i",
@@ -346,6 +352,10 @@
         },
         {
           "command": "vscode-db2i.jobManager.editJobProps",
+          "when": "never"
+        },
+        {
+          "command": "vscode-db2i.jobManager.editSelfCodes",
           "when": "never"
         },
         {
@@ -516,6 +526,11 @@
         },
         {
           "command": "vscode-db2i.jobManager.editJobProps",
+          "when": "view == jobManager && viewItem == sqlJob",
+          "group": "inline"
+        },
+        {
+          "command": "vscode-db2i.jobManager.editSelfCodes",
           "when": "view == jobManager && viewItem == sqlJob",
           "group": "inline"
         },
--- a/package.json
+++ b/package.json
@@ -291,7 +291,7 @@
       },
       {
         "command": "vscode-db2i.jobManager.editSelfCodes",
-        "title": "Edit SELFCODES",
+        "title": "Edit SELF codes",
         "category": "Db2 for i",
         "icon": "$(bracket-error)"
       },
@@ -319,7 +319,7 @@
       },
       {
         "command": "vscode-db2i.jobManager.getSelfErrors",
-        "title": "Get SELFCODES Errors",
+        "title": "Get SELF codes Errors",
         "category": "Db2 for i"
       },
       {
--- a/package.json
+++ b/package.json
@@ -541,7 +541,7 @@
         {
           "command": "vscode-db2i.jobManager.editSelfCodes",
           "when": "view == jobManager && viewItem == sqlJob",
-          "group": "inline"
+          "group": "self@2"
         },
         {
           "command": "vscode-db2i.jobManager.copyJobId",
@@ -566,7 +566,7 @@
         {
           "command": "vscode-db2i.jobManager.getSelfErrors",
           "when": "view == jobManager && viewItem == sqlJob",
-          "group": "trace@3"
+          "group": "self@1"
         },
         {
           "command": "vscode-db2i.jobManager.newConfig",
--- a/src/testing/manager.ts
+++ b/src/testing/manager.ts
@@ -91,24 +91,6 @@ export const ManagerSuite: TestSuite = {
       assert.strictEqual(JobManager.getSelection().name, runningJobs[0].name);
 
       await JobManager.endAll();
-    }},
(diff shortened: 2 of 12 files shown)


ajshedivy:
Added testing framework for self codes. We can add more as we please

worksofliam (commented):
Initial reaction is good. I am currently looking for a system to do tests on (my 7.3 box won't cut it).

In the mean time, the first thing to change is any reference of "SELFCODES" in labels to be "SELF codes"/"SELF code".

src/views/jobManager/selfCodes/selfCodesBrowser.ts
```diff
@@ -0,0 +1,24 @@
...
+  description?: string;
+  detail?: string;
+
+  constructor(object: SelfCodeObject) {
+    this.label = object.code;
+    this.description = object.message;
+  }
+}
+
+export async function setSelfCodes(codes: string[]) {
```
worksofliam, line 16: This function should actually be an instance method in `sqlJob`.

src/views/jobManager/jobManagerView.ts
```diff
@@ -127,6 +129,76 @@ export class JobManagerView implements TreeDataProvider<any> {
...
+            ];
+
+            quickPick.selectedItems = currentSelfCodeItems;
+
+            quickPick.onDidAccept(async () => {
+              const selections = quickPick.selectedItems;
+              // SET SYSIBMADM.SELFCODES = SYSIBMADM.VALIDATE_SELF('-514, -204, -501, +30, -199');
+              if (selections) {
+                const codes: string[] = selections.map((code) => code.label);
+                setSelfCodes(codes);
```
worksofliam, line 174: Here we should be calling an instance method on the selected job (`selected`) to set the SELF codes.

package.json
```diff
@@ -289,6 +289,12 @@
         "category": "Db2 for i",
         "icon": "$(edit)"
       },
+      {
+        "command": "vscode-db2i.jobManager.editSelfCodes",
+        "title": "Edit SELFCODES",
```
worksofliam, line 294: We need to fix all label references of "SELFCODES" to "SELF codes"

worksofliam (changes requested):

worksofliam (changes requested):
Initial review is that it works, but the organisation of the 'set codes' method needs to be part of the SQL job and not a seperate function.

worksofliam (changes requested):
Initial review is that it works, but the organisation of the 'set codes' method needs to be part of the SQL job and not a seperate function.

ajshedivy:
current testing state:

image

src/testing/manager.ts
```diff
@@ -2,7 +2,8 @@ import assert from "assert";
 import { TestSuite } from ".";
 import { JobManager } from "../config";
 import { ServerComponent } from "../connection/serverComponent";
-import { JobStatus } from "../connection/sqlJob";
+import { JobStatus, SQLJob } from "../connection/sqlJob";
+import { testSelfCodes } from "../views/jobManager/selfCodes/selfCodesTest";
```
worksofliam, line 6: Used import here. Should be removed.

package.json
```diff
@@ -539,6 +563,11 @@
           "when": "view == jobManager && viewItem == sqlJob",
           "group": "trace@2"
         },
+        {
+          "command": "vscode-db2i.jobManager.getSelfErrors",
+          "when": "view == jobManager && viewItem == sqlJob",
+          "group": "trace@3"
```
worksofliam, line 569: I think this belongs in a new group, specifically for SELF. Perhaps:

```
"group": "self@1"
```

package.json
```diff
@@ -519,6 +538,11 @@
           "when": "view == jobManager && viewItem == sqlJob",
           "group": "inline"
         },
+        {
+          "command": "vscode-db2i.jobManager.editSelfCodes",
+          "when": "view == jobManager && viewItem == sqlJob",
+          "group": "inline"
```
worksofliam, line 544: I don't think this should be inline right now, since we already had 4 icons. I think the group should be in the SELF group. Perhaps:

```
"group": "self@2"
```

worksofliam (changes requested):
Code looks good - just a couple of minor `package.json` changes, and perhaps some changing of the test cases is in order.

I actually think the SELF support should go into its own `TestSuite` instead of being part of the manager. Also, only some of the test cases are passing. What am I doing wrong? I am on a a 7.4 system. (`oss74dev`)

image

ajshedivy:
> Code looks good - just a couple of minor `package.json` changes, and perhaps some changing of the test cases is in order.
> 
> I actually think the SELF support should go into its own `TestSuite` instead of being part of the manager. Also, only some of the test cases are passing. What am I doing wrong? I am on a a 7.4 system. (`oss74dev`)
> image

hmm, interesting the tests look okay for me:

image

worksofliam (approved):
Thanks @ajshedivy - the tests are passing! Let's merge.

image

PR #158, opened 2023-10-17
Feature/local server component

### LINKED ISSUE #157: Install Server Component from local without fetching from GitHub

worksofliam:
Packages a specific version of the Server Component into the extension vsix, which will then be automatically installed when connected to a server.

Closes #157.

### DIFF
--- a/src/connection/serverComponent.ts
+++ b/src/connection/serverComponent.ts
@@ -9,8 +9,6 @@ import { OutputChannel, extensions, window } from "vscode";
 import { stat } from "fs/promises";
 import { SERVER_VERSION_FILE } from "./SCVersion";
 
-const octokit = new Octokit();
-
 const ExecutablePathDir = `$HOME/.vscode/`;
 
 export enum UpdateStatus {
--- a/src/connection/serverComponent.ts
+++ b/src/connection/serverComponent.ts
@@ -11,9 +11,6 @@ import { SERVER_VERSION_FILE } from "./SCVersion";
 
 const octokit = new Octokit();
 
-// During development, you can set the SERVER_VERSION in .vscode/launch.json
-// Otherwise, fall back to the working version
-
 const ExecutablePathDir = `$HOME/.vscode/`;
 
 export enum UpdateStatus {
--- a/src/connection/serverComponent.ts
+++ b/src/connection/serverComponent.ts
@@ -1,20 +1,18 @@
 import { getInstance } from "../base";
 
 import { Octokit } from "octokit";
-import fetch from 'node-fetch';
 
 import { Config } from "../config";
 import path from "path";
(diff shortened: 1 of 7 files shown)


src/connection/serverComponent.ts
```diff
@@ -1,20 +1,18 @@
...
 
-import { writeFile, unlink } from "fs/promises";
-import os from "os";
+import { stat } from "fs/promises";
+import { SERVER_VERSION_FILE } from "./SCVersion";
 
 const octokit = new Octokit();
 
 // During development, you can set the SERVER_VERSION in .vscode/launch.json
 // Otherwise, fall back to the working version
```
william-xiang, line 15: I guess these comments can be removed now?

src/extension.ts
```diff
@@ -28,6 +28,8 @@ export interface Db2i {
 
 export function activate(context: vscode.ExtensionContext): Db2i {
 
+  console.log(context.extensionPath);
```
william-xiang, line 31: Not sure if this is added on purpose, seems it's for debugging only.

src/connection/serverComponent.ts
```diff
@@ -90,79 +88,54 @@ export class ServerComponent {
...
 
-        if (lastInstalledName !== basename || this.installed === false) {
-          const updateQuestion = await window.showInformationMessage(
-            `Database Extension update required`, 
-            {
-              modal: true,
-              detail: `An update to the database server component is required: ${basename}`
-            }, 
-            `Update`
-          );
```
william-xiang, line 125: So here we will always update the server component if there is update available without letting user decide?

src/connection/serverComponent.ts
```diff
@@ -90,79 +88,54 @@ export class ServerComponent {
...
+            command: `echo ${ExecutablePathDir}`
+          });
 
-              await connection.uploadFiles([{local: tempFile, remote: remotePath}]);
+          if (commandResult.code === 0 && commandResult.stderr === ``) {
+            const remotePath = path.posix.join(commandResult.stdout, basename);
 
-              // Clean up the temp file
-              unlink(tempFile);
+            ServerComponent.writeOutput(JSON.stringify({remotePath, ExecutablePathDir}));
```
william-xiang, line 120: I think it's better to use the actual path value from commandResult.

william-xiang (changes requested):
The other two question/suggestion:
1. should the version be bumped for this change?
2. exclude the dsc.ts from the bundle

edmundreinhardt (approved):

william-xiang (approved):
Approved. Thank you.

worksofliam:
@william-xiang 

1. We won't do a version bump here because we do not need to.
2. I removed `dsc.ts` from the vsix build.

PR #161, opened 2023-10-28
Better support for finding statement type when EXEC is used

worksofliam:
Fixes bug where `INTO` clause was being removed during `INSERT` statement.


worksofliam:
Tested by @forstie

forstie:
Fix works well, thanks!

PR #164, opened 2023-11-08
Update SQL formatter and allow SQL keyword in lower- or uppercase

chrjorgensen:
This PR will
- update the [SQL formatter](https://github.com/sql-formatter-org/sql-formatter) to the current version of 13.1.0
- use the new language `db2i` implemented in SQL Formatter version 13.1.0
- add a configuration value `sqlFormat.keywordCase` to allow the user to have SQL keywords in lower- or uppercase

With this PR and a configuration setting of `lower`, the generated SQL for `QGPL.QAIDRKEYS` will be like this:

image

Before this would be formatted like this:

image

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12,7 +12,7 @@
         "lru-cache": "^6.0.0",
         "node-fetch": "^3.3.1",
         "octokit": "^2.0.14",
-        "sql-formatter": "^4.0.2"
+        "sql-formatter": "^12.2.4"
       },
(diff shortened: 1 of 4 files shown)


worksofliam:
Lower case keywords is soooooo nice. Thank you, @chrjorgensen, and thank you even more for contributing on the formatter.

worksofliam (approved):
A very small PR that is not high impact. Thanks for your work here.

PR #165, opened 2023-11-10
Initial work on Notebooks

worksofliam:
* [x] Initial port from vscode-ibmi-notebooks
* [x] Ability to open existing .sql sources as a new Notebook, separating statements into separate cells (SQL files have a button to open as a notebook)
* [x] Consider how we can generate different formats of data other than tables for relational data (we added charts!)
* [x] Add chart width (`max-width`?) settings to extension
* [x] Ability to create a new empty Notebook from the SQL Job Manager view, similar to how we have a 'new empty file' button
* [x] New Notebook examples under the Examples view (API has been added `isNotebook`)
* [ ] ~Reset result set view when cell is executed~
* [x] Comma at the end of statements supported
* [x] Docs on Notebooks - requires dev work to be finished


src/notebooks/IBMiSerializer.ts
```diff
@@ -0,0 +1,64 @@
...
+
+interface RawNotebookCell {
+  language: string;
+  value: string;
+  kind: vscode.NotebookCellKind;
+}
+
+export function notebookInit() {
+  const openNotebook = vscode.commands.registerCommand(`vscode-db2i.openNotebook`, (node) => {
+    const uri = node ? node.resourceUri : vscode.Uri.parse(`untitled:` + `notebook.inb`);
```
davecharron, line 14: Should there be a counter here that could be used to generate the file name?  As it is now, with the default file name `notebook.inb`, only one unsaved notebook can be open at a time.

worksofliam: Fixed in a new commit. Changed the code to use a better VS Code API to open blank documents, which handles the title like this for us.

forstie (approved):
tip top

forstie (approved):
rockin

PR #166, opened 2023-11-10
Add new format provider

worksofliam:
Since we have a nicer SQL formatter now, I have decided to add a format provider which enables the standard 'Format Document' to work with SQL sources.


worksofliam:
@chrjorgensen We'll have `sqlformat.json` soon 🤣

PR #168, opened 2023-11-16
Feature/self codes p2

ajshedivy:
**Part 2**

- [x] Ability to store self codes into config to they are set when the job starts up
- [x] If self codes are set, then refresh the error log after every statement into a new view (like a panel at the bottom)
- [x] Format JSON in result set (https://github.com/codefori/vscode-db2i/blob/7c3126cfc7f4d0f0b07f7580dc021c04044fd67f/src/views/results/html.ts#L151)


src/views/selfCodesPanel/index.ts
```diff
@@ -0,0 +1,12 @@
+import * as vscode from 'vscode';
+
+class SelfCodesPanelProvider {
```
worksofliam, line 3: Is this used anywhere? There is no `export` in this file.

src/views/results/index.ts
```diff
@@ -1,18 +1,19 @@
...
 import { changedCache } from "../../language/providers/completionItemCache";
-import { IRange, ObjectRef, ParsedEmbeddedStatement, StatementGroup, StatementType } from "../../language/sql/types";
+import Document from "../../language/sql/document";
+import { ObjectRef, ParsedEmbeddedStatement, StatementGroup, StatementType } from "../../language/sql/types";
+import { updateStatusBar } from "../jobManager/statusBar";
+import * as html from "./html";
 import Statement from "../../language/sql/statement";
-
+import { SQLJob } from "../../connection/sqlJob";
+import LRUCache from "lru-cache";
```
worksofliam, line 16: A few unused imports here. These should be removed.

src/views/results/html.ts
```diff
@@ -141,6 +141,15 @@ export function generateScroller(basicSelect: string, isCL: boolean): string {
             });
           }
 
+          function isJsonString(str) {
```
worksofliam, line 144: I think this `isJsonString` should be converted to be more like `getFormattedString`, where we could pass JSON, XML, etc, later on, and have it return the correct format. Then, in the catch, return the original `str`. Also, `str` here is missing the type on the parameter.

src/views/results/html.ts
```diff
@@ -179,4 +196,116 @@ export function generateScroller(basicSelect: string, isCL: boolean): string {
...
+              // Insert a cell at the end of the row
+              var newCell = newRow.insertCell();
+
+              // Append a text node to the cell
+              var newText = document.createTextNode(colName);
+              newCell.appendChild(newText);
+            });
+          }
+
+          function isJsonString(str) {
```
worksofliam, line 260: Same here for `isJsonString`

src/connection/sqlJob.ts
```diff
@@ -4,6 +4,7 @@ import { ServerComponent } from "./serverComponent";
 import { JDBCOptions, ConnectionResult, Rows, QueryResult, JobLogEntry, CLCommandResult, VersionCheckResult, GetTraceDataResult, ServerTraceDest, ServerTraceLevel, SetConfigResult, QueryOptions } from "./types";
 import { Query } from "./query";
 import { EventEmitter } from "stream";
+import { commands } from "vscode";
```
worksofliam, line 7: `vscode` should not be imported in this class as we will make it reusable outside of VS Code. Sorry!

worksofliam (changes requested):

src/views/selfCodesPanel/index.ts
ajshedivy: just used later in `vscode-db2i.runEditorStatement`

src/views/results/index.ts
```diff
@@ -253,6 +320,22 @@ export function initialise(context: vscode.ExtensionContext) {
...
+                    await selfCodeErrorProvider.setTableData(data);
+                    selfCodeErrorProvider.selfCodeCache = data.length;
+                  }
+                } else {
+                  await vscode.commands.executeCommand(`setContext`, `vscode-db2i:selfCodeCountChanged`, false);
+                }
+              }
+              
               if (statementDetail.qualifier === `statement` && statementDetail.history !== false) {
                 vscode.commands.executeCommand(`vscode-db2i.queryHistory.prepend`, statementDetail.content);
```
worksofliam, line 340: We should not prepend the SQL statement to fetch from the error log.

src/views/results/index.ts
```diff
@@ -116,6 +115,67 @@ class ResultSetPanelProvider {
       queryId: ``
     });
   }
+  setError(error) {
+    // TODO: pretty error
+    this._view.webview.html = `<p>${error}</p>`;
+  }
+}
+
+class SelfCodePanelProvider {
```
worksofliam, line 124: This should be in a seperate file.

src/views/results/index.ts
```diff
@@ -142,6 +142,40 @@ export interface ParsedStatementInfo extends StatementInfo {
...
+      webviewOptions : {retainContextWhenHidden: true}
+    }),
+    vscode.commands.registerCommand(`vscode-db2i.selfCodeErrorPanel.fetch`, async (job: SQLJob, showErrors: boolean) => {
+      await selfCodeErrorProvider.ensureActivation();
+      const content = `SELECT * FROM QSYS2.SQL_ERROR_LOG WHERE JOB_NAME = '${job.id}'`;
+      let prevDataLength = job.selfCodeCounter;
+      let updateView: boolean = false;
+      let data = [];
+      for (let count = 0; count < 5; count++) {
+        const curData = await JobManager.runSQL(content, undefined);
```
worksofliam, line 158: Instead of using the `JobManager` here, you should just use the attached `job` parameter (`job.query`)

src/connection/sqlJob.ts
worksofliam: This is still requred ^

PR #172, opened 2023-11-24
Add SQL identifier case - new in SQL formatter 14.0.0

chrjorgensen:
This PR will
- update SQL formatter to version 14.0.0, which includes the new option for identifier formatting
- add a configuration value `sqlFormat.identifierCase` to allow the user to have SQL identifier in lower- or uppercase or not changed. Default is `preserve` to not change the identifier case unintentionally.

With this support the following SQL statement

image

will be formatted like this (with identifiers set to uppercase and keywords to lowercase):

image

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12,7 +12,7 @@
         "lru-cache": "^6.0.0",
         "node-fetch": "^3.3.1",
         "octokit": "^2.0.14",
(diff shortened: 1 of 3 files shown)


worksofliam (approved):
New config is awesome! `preserve` is a nice touch.

Thank you @chrjorgensen!

PR #184, opened 2023-12-21
Don't use variant characters in SQL statement

chrjorgensen:
This PR changes the use of `||` in SQL statement to `concat` - the `|` is a variant character...


worksofliam:
Tested and works good! Thanks @chrjorgensen!!

PR #185, opened 2024-01-07
Examples browser cleanup

sebjulliand:
I noticed that the Examples browser looped when loading if no SQL job was started, because loading the Services called `refresh` which in turn called `getChildren`...and if the Services are still undefined (because no SQL job), then it loops.

So this PR fixes that. Loading the Services example doesn't call `refresh` anymore (it wasn't needed in the first place), and an information item is displayed under the Service folder to ask for an SQL job.

A `Reload` was added to the browser to allow to reload the Services example if needed.

The Examples are now sorted alphabetically and their icon is the SQL file's.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "vscode-db2i",
-  "version": "0.5.2",
+  "version": "0.6.0",
   "lockfileVersion": 2,
   "requires": true,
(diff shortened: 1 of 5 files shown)


worksofliam (approved):
Cool - this logic is much easier to follow. Thanks @sebjulliand!

PR #186, opened 2024-01-07
Build/watchTask

sebjulliand:
- Removed ts-loader as it's not used
- Added es-build problem matcher to the extension recommendations
- Added a watch task and made it the default when debugging


worksofliam (approved):
This is a great cleanup. Thanks!

worksofliam (approved):
This is a great cleanup. Thanks!

PR #195, opened 2024-02-08
Feature/cancel

worksofliam:
Shows cancel button when running an SQL statement.

* [x] Cancels the statement when switching job in the Job Manager
* [x] Cancel button is shown during statement execution
* [x] Cancel button is shown during explain execution
* [x] Cancel button is shown during `csv`, `json`, etc, execution
* [ ] ~Running a statement while one is already running will cancel the previous statement~ <- we're not doing that now
* [ ] **Optional**: updating the SQL Job node based on the `busy` state

### DIFF
--- a/src/connection/sqlJob.ts
+++ b/src/connection/sqlJob.ts
@@ -1,4 +1,3 @@
-import { CommandResult } from "@halcyontech/vscode-ibmi-types";
 import { getInstance } from "../base";
 import { ServerComponent } from "./serverComponent";
 import { JDBCOptions, ConnectionResult, Rows, QueryResult, JobLogEntry, CLCommandResult, VersionCheckResult, GetTraceDataResult, ServerTraceDest, ServerTraceLevel, SetConfigResult, QueryOptions } from "./types";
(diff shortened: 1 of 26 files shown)


davecharron:
The explain task is more complicated than simply running a statement, so canceling would necessarily be different.
The server component encapsulates the `DoVe` request, which includes multiple steps before and after the actual SQL statement execution.  Canceling would involve potentially resetting QAQQINI options, ending/deleting the monitor, and, if the current step is at the point of executing the SQL statement, canceling it.
It would seem that to accomplish that, a server component change would be necessary to provide a way to cancel a request.  Likely, the base `ClientRequest` class would get a new `cancel()` method.  `DoVe` would override it to do the gymnastics described above.  For other request types, we could either have the `cancel()` method in the base class no-op, or we could push up the cancel logic and implement in the other request sub-classes, ie. `RunSql`.

src/connection/sqlJob.ts
```diff
@@ -140,7 +137,9 @@ export class SQLJob {
   }
 
   getStatus() {
-    return this.status;
+    const currentListnerCount = this.responseEmitter.eventNames().length;
```
davecharron, line 140:
```suggestion
    const currentListenerCount = this.responseEmitter.eventNames().length;
```

davecharron (approved):
I'm not seeing any blockers to merging.

src/connection/sqlJob.ts
worksofliam: Added in new commit

PR #196, opened 2024-02-13
Move advice generation to other file

worksofliam:
Simply moves the logic that generates the advice script content to another file to slim down the result set file size.


davecharron (approved):
I like it, looks good.

PR #205, opened 2024-03-05
Use `with` statement for filtering to work

worksofliam:
* Wrap `select` statements for multiple object types in a `with` clause to enable schema filtering to work again


worksofliam:
New test case is passing 

image

davecharron (approved):
Looks good, kudos on the new test.

PR #206, opened 2024-03-06
Change colour of ship base

worksofliam:
When the theme was Light+, the ship was blended into the background.


davecharron (changes requested):

worksofliam:
Much better. Thanks @davecharron

davecharron (approved):
Looks good in both light and dark modes.

Issue #137, opened 2023-09-22
Feature: Self Codes


worksofliam:
**Part 1**

- [x] Ability to set self codes on a job
- [x] Ability to fetch error log for a job

**Part 2**

```sql
SELECT user_name, logged_time, logged_sqlstate, logged_sqlcode, matches, stmttext, 
       message_text, message_second_level_text 
  FROM qsys2.sql_error_log, lateral 
    (select * from TABLE(SYSTOOLS.SQLCODE_INFO(logged_sqlcode)))
where user_name = current_user
order by logged_time desc
```

* [x] New tree view for SELF logs
  * [x] Can only be shown when we have an SQL job running
  * [x] This view needs a button to clear the self code logs for the current user
  * [x] Manually refresh button
  * [x] Node style matches the following:

```
  // 01534 (181) - 15 hits, Values in date, time, or timestamp string not valid.
    // Hover: STMTTEXT \n\n---\n\nSECOND_LEVEL_TEXT
```

* [x] SQL job needs a option to set the level
  * [x] No longer setting by specific code
  * [x] Setting by: *ALL, *WARN, *ERROR, *NONE

```sql
set sysibmadm.selfcodes = '*ALL';  -- or *ERROR or *WARN or *NONE
```

* [x] Interval to fetch the SELF log
  1. [x] Run the provided statement to get the logs for the current user
  2. [x] Update the SELF log tree view with the new nodes based on the log data
* [x] We need a way to turn on/off the interval, which also means turning off and always hiding the view
* [ ] Ensure SELF Codes views and buttons only work on 7.4+
   * We have the version in `OSData`, we can just set a context value to enable/disable based on that
* Use decoration provider for hit count in SELF Code view

---

Look into how to incorporate this?

```
select a.*
  from QSYS2.SQL_ERROR_LOG l, lateral (
         select *
           from json_table(INITIAL_STACK, 'lax $.initial_stack[*]'
               columns(
                 ORDINAL_POSITION integer path 'lax $."ORD"',
                 PROGRAM_TYPE varchar(10) ccsid 37 path 'lax $."TYPE"',
                 PROGRAM_LIB  varchar(10) ccsid 37 path 'lax $."LIB"',
                 PROGRAM_NAME  varchar(10) ccsid 37 path 'lax $."PGM"',
                 MODULE_NAME  varchar(10) ccsid 37 path 'lax $."MODULE"',
[...]

PR #207, opened 2024-03-08
Feature/new self codes

ajshedivy:
### Issues:

 - #137 

#### TODO

- [x] better check for os version and ptf levels, see https://github.com/codefori/vscode-db2i/blob/c1e7da1ed85fa242844889bb74b0e5867607461b/src/config.ts#L21 (Liam - met with Scott, he said current logic should work ok)
- [x] Self Code default enablement (Liam)
- [x] Use decoration provider for hit count in SELF Code view instead of in the node description (we already do this for Dove written by Dave!)
- [x] Configuration option to have auto-refresh enabled/disabled automatically
- [x] Make use of `viewsWelcome` in `package.json` for when there are no SELF code errors.
- [x] Fix SQL statement to clear SELF logs: 
image
- [x] Button on the SELF error to copy the SQL statement to the clipboard
- [x] Show reason code when it's not null
- [x] Update SELF code tests
- [x] Change titles of SELF quick pick and view to full name
    * Quick pick: 'Set logging level for SELF'
    * View: 'SQL Error Log Facility (SELF)'
- [x] Add button in SELF view to open the docs: https://www.ibm.com/docs/en/i/7.5?topic=tools-sql-error-logging-facility-self

#### Things to test:

- [x] Ensure self codes are stored into saved settings
- [x] Ensure when starting a job from configuration, SELF codes from the settings are applied automatically
- [x] Works under *PGMR auth on the user profile (e.g. not *ALLOBJ) - scott to test


src/connection/sqlJob.ts
```diff
@@ -274,10 +274,9 @@ export class SQLJob {
     return rpy;
   }
 
-  async setSelfCodes(codes: string[]) {
-    const signedCodes: String[] = codes.map(code => [code, `-${code}`]).flat();
+  async setSelfCodes(codes: string) {
     try {
-      const query: string = `SET SYSIBMADM.SELFCODES = SYSIBMADM.VALIDATE_SELF('${signedCodes.join(', ')}')`
+      const query: string = `SET SYSIBMADM.SELFCODES = '${codes}'`
       await this.query<any>(query).run();
```
worksofliam, line 280: Consider using the second parameter to `query`, which accepts binding parameters in place of `?` in the statement here.

src/views/jobManager/jobManagerView.ts
```diff
@@ -185,20 +190,29 @@ export class JobManagerView implements TreeDataProvider<any> {
...
-          const content = `SELECT * FROM QSYS2.SQL_ERROR_LOG WHERE JOB_NAME = '${selected.job.id}'`;
-
-          vscode.commands.executeCommand(`vscode-db2i.runEditorStatement`, {
-            content,
-            qualifier: `statement`,
-            open: false,
-          });
-        }
-      }),
+      // vscode.commands.registerCommand(`vscode-db2i.jobManager.getSelfErrors`, async (node?: SQLJobItem) => {
```
worksofliam, line 193: Remove this large comment block of code. It will be in history if we need it

worksofliam (changes requested):
Just a couple of small bits. Looks good so far. I will update your todo list with a couple more items that I want to tackle here.

src/views/jobManager/selfCodes/selfCodesResultsView.ts
```diff
@@ -0,0 +1,155 @@
...
+
+  getTreeItem(element: any): TreeItem | Thenable<TreeItem> {
+    return element;
+  }
+  async getChildren(element?: any): Promise<any[]> {
+    if (element) {
+      return [];
+    } else {
+      const selfCodes = await this.getSelfCodes();
+      return selfCodes.map((error) => {
```
worksofliam, line 84: There is a chance that `selfCodes` can be undefined here, as `getSelfCodes` can return `undefined` if no job is running. Considering returning an empty array so then no items will show up, and then use `viewsWelcome` in the `package.json` file to have default text.

worksofliam: I added the if statement to make sure it is defined!

src/views/jobManager/selfCodes/selfCodesResultsView.ts
```diff
@@ -0,0 +1,179 @@
...
+    public readonly details: string,
+    public readonly hoverMessage: vscode.MarkdownString,
+    public readonly collapsibleState: vscode.TreeItemCollapsibleState,
+    error: SelfCodeNode
+  ) {
+    super(errorMessage, collapsibleState);
+    this.tooltip = hoverMessage; // Hover text
+    this.tooltip.isTrusted = true;
+    this.description = details; // Additional details shown in the tree view
+    this.resourceUri = vscode.Uri.parse(`selfCodeTreeView:${encodeURIComponent(error.MATCHES.toString())}`);
```
worksofliam, line 145: Let's use `Uri.from` here instead of `parse`

src/extension.ts
```diff
@@ -34,6 +36,10 @@ export function activate(context: vscode.ExtensionContext): Db2i {
 
   loadBase();
 
+  const exampleBrowser = new ExampleBrowser(context);
+  const selfCodesView = new selfCodesResultsView(context);
+  const selfCodeFileDecorator = new SelfTreeDecorationProvider();
```
worksofliam, line 41: Add this to the `context.subscriptions`

src/connection/sqlJob.ts
worksofliam: We spoke with Jesse. There is a bug in the SC that stops this from working.

worksofliam (approved):
We have worked with Scott and he is happy. I have done a final code review, fixed some minor things, and it's looking sharp. Let's merge.

forstie:
Thanks for all the revisions.
Finally, SELF is coming center stage.

PR #211, opened 2024-03-19
Fix issue where reconnected job had a bad status

worksofliam:
Previously, when editing job settings, the new job would be successfully created. But, it was still in `ended` state. This solves that problem so job settings can be changed correctly.

### DIFF
--- a/src/connection/sqlJob.ts
+++ b/src/connection/sqlJob.ts
@@ -181,19 +181,19 @@ export class SQLJob {
 
     const connectResult: ConnectionResult = JSON.parse(result);
 
-    if (connectResult.success !== true) {
+    if (connectResult.success === true) {
+      this.status = JobStatus.Ready;
+    } else {
       this.dispose();
       this.status = JobStatus.NotStarted;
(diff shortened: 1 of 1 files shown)


src/connection/sqlJob.ts
```diff
@@ -181,19 +181,19 @@ export class SQLJob {
...
+    if (connectResult.success === true) {
+      this.status = JobStatus.Ready;
+    } else {
       this.dispose();
       this.status = JobStatus.NotStarted;
       throw new Error(connectResult.error || `Failed to connect to server.`);
     }
 
-    if (this.status === JobStatus.Ended) {
+    if (this.status !== JobStatus.Ready) {
```
davecharron, line 192: Is this even necessary?  It doesn't seem like this could ever happen.  In the `if-else` above, status is either set to `JobStatus.Ready`, or we dispose and throw an `Error`.

worksofliam: True story!

davecharron (approved):

PR #212, opened 2024-03-20
List query history by date

worksofliam:
* Change how we store history internally (no longer a list of strings)
* Group statements by simple period (past day, past week, past month, later)
* All but first are collapsed by default

warning: if you test this PR, it will break your existing query history

### DIFF
--- a/src/views/queryHistoryView.ts
+++ b/src/views/queryHistoryView.ts
@@ -84,19 +84,50 @@ export class queryHistory implements TreeDataProvider<any> {
         return timePeriod.getChildren();
 
       } else {
-        let currentList = Config.getPastQueries();
-        const timePeriods: { [key: string]: PastQueryNode[] } = {};
+        const currentList = Config.getPastQueries();
+        
+        const day = 60 * 60 * 24;
+        const week = day * 7;
+        const month = day * 30;
+        
+        const now = Math.floor(Date.now() / 1000);
+        const dayAgo = now - day;
+        const weekAgo = now - week;
+        const monthAgo = now - month;
 
-        for (let queryItem of currentList) {
-          const date = new Date(queryItem.unix * 1000);
-          const period = date.toDateString();
-          if (!timePeriods[period]) {
-            timePeriods[period] = [];
+        let dayQueries: PastQueryNode[] = [];
+        let weekQueries: PastQueryNode[] = [];
+        let monthQueries: PastQueryNode[] = [];
+        let laterQueries: PastQueryNode[] = [];
+
+        currentList.forEach(queryItem => {
+          if (queryItem.unix < dayAgo) {
+            dayQueries.push(new PastQueryNode(queryItem.query));
+          } else if (queryItem.unix < weekAgo) {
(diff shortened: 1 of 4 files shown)


davecharron:
Should we consider generic group captions as the dates become more removed from the present?  Something like what is done in Outlook, for example.
image

worksofliam:
@davecharron I will work on that.

src/views/queryHistoryView.ts
```diff
@@ -72,17 +78,79 @@ export class queryHistory implements TreeDataProvider<any> {
...
+          nodes.push(new TimePeriodNode(`Past day`, dayQueries, true));
+        }
+        if (weekQueries.length > 0) {
+          nodes.push(new TimePeriodNode(`Past week`, weekQueries));
+        }
+        if (monthQueries.length > 0) {
+          nodes.push(new TimePeriodNode(`Past month`, monthQueries));
+        }
+        if (laterQueries.length > 0) {
+          nodes.push(new TimePeriodNode(`Later`, laterQueries));
```
davecharron, line 127:
```suggestion
          nodes.push(new TimePeriodNode(`Older`, laterQueries));
```

davecharron (changes requested):

src/views/queryHistoryView.ts
```diff
@@ -72,17 +78,79 @@ export class queryHistory implements TreeDataProvider<any> {
...
+        currentList.forEach(queryItem => {
+          if (queryItem.unix < dayAgo) {
+            dayQueries.push(new PastQueryNode(queryItem.query));
+          } else if (queryItem.unix < weekAgo) {
+            weekQueries.push(new PastQueryNode(queryItem.query));
+          } else if (queryItem.unix < monthAgo) {
+            monthQueries.push(new PastQueryNode(queryItem.query));
+          } else {
+            laterQueries.push(new PastQueryNode(queryItem.query));
+          }
```
davecharron, line 112:
```suggestion
          if (queryItem.unix > monthAgo) {
             laterQueries.push(new PastQueryNode(queryItem.query));
          } else if (queryItem.unix > weekAgo) {
            monthQueries.push(new PastQueryNode(queryItem.query));
         } else if (queryItem.unix > dayAgo) {
            weekQueries.push(new PastQueryNode(queryItem.query));
         } else {
            dayQueries.push(new PastQueryNode(queryItem.query));
         }
```

davecharron (changes requested):
Try this order

davecharron:
If a query from a category other than _Past day_ is run, it gets added to the _Past day_ category, but does not get removed from the previous category.

davecharron:
> If a query from a category other than _Past day_ is run, it gets added to the _Past day_ category, but does not **always** get removed from the previous category. I'll step through the failing case to see what the deal is.

Looks like it's a weirdness where the statement ended up in two categories.  Not sure how, but could be related to the warning on this PR that testing it breaks existing history.  After manually deleting one of the instances, it behaves as expected.

davecharron (approved):

PR #213, opened 2024-03-20
Add calls to change permissions of server component

### LINKED ISSUE #76: Install server component with correct permissions
Server component is installed with public read authority, perhaps should/could be public *EXCLUDE

`$HOME/.vscode` has permissions
```ada
d rwx r-s r-x
```
(perhaps inherited from home dir)

No glaring security hole here, but we should: 
- [ ] determine the proper authorities that should be set
- [ ] change the code to set them accordingly (if applicable)

worksofliam:
Fixes #76

> Server component is installed with public read authority, perhaps should/could be public *EXCLUDE

### DIFF
--- a/src/connection/serverComponent.ts
+++ b/src/connection/serverComponent.ts
@@ -112,7 +112,19 @@ export class ServerComponent {
 
             ServerComponent.writeOutput(JSON.stringify({remotePath, ExecutablePathDir}));
 
+            const createDirResult = await connection.sendCommand({
+              command: `mkdir -p ${ExecutablePathDir} && chmod 500 ${ExecutablePathDir}`
+            });
+
+            this.writeOutput(JSON.stringify(createDirResult));
+
             await connection.uploadFiles([{local: assetPath, remote: remotePath}]);
+
+            const scAuth = await connection.sendCommand({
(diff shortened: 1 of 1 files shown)


src/connection/serverComponent.ts
```diff
@@ -112,8 +112,20 @@ export class ServerComponent {
 
             ServerComponent.writeOutput(JSON.stringify({remotePath, ExecutablePathDir}));
 
+            const createDirResult = await connection.sendCommand({
+              command: `mkdir -p ${ExecutablePathDir} && chmod 500 ${ExecutablePathDir}`
```
davecharron, line 116: Why not 700?  Doesn't the lack of write permission prevent uploading the jar?

worksofliam: I believe not since every user uploads the server component to their home directory, so in theory they own it.

@ThePrez care to chime in?

davecharron: I ran a test using `chmod 500` and I could still upload and write to a new file in the directory.  So, I guess it's ok.

davecharron (approved):
Changes look reasonable to me.

PR #215, opened 2024-03-25
Fix resultSet view not grabing focus after reconnection

sebjulliand:
This PR fixes an issue causing the ResultSet view to not grab the focus when running a statement after disconnecting and reconnecting to an IBM i.

This was caused by the `ResultSetPanelProvider` not keeping track of the `disposed` status of the webview, preventing the `ensureActivation` from working properly.

By resetting the reference to the resolved webview once it's disposed, the `ensureActivation` method can work as expected after reconnecting to a system.

## How to test
1. Connect to an IBM i
2. Run a statement
3. The Result Set webview must grab the focus and display the results
4. Disconnect
5. Connect again
6. Run a statement
7. The Result Set webview must grab the focus and display the results again

Without this PR, step `7.` doesn't happen and the user must manually reveal the result set webview.


davecharron (approved):
Changes look reasonable, seems to work as expected.

Issue #102, opened 2023-08-01
SQL results view should include which job the statement was run in


ThePrez:
Especially when there could be multiple SQL jobs in use, it would be good to get some visual indicator as to which job a query or SQL statement was run in

PR #217, opened 2024-03-25
Add footer to results table so status is always visible

davecharron:
The results table view features a span that is updated with status, eg. statement running, row count...
The problem is that with a larger result set, the span remains out of view at the bottom until all rows have been retrieved and the result table scrolled to the end.
This PR fixes the issue by adding a footer to the table where the status is always displayed.

### DIFF
--- a/src/views/html.ts
+++ b/src/views/html.ts
@@ -20,6 +20,13 @@ export function getHeader(options: {withCollapsed?: boolean} = {}): string {
       top: 0;           /* Don't forget this, required for the stickiness */
     }
 
+    tfoot tr {
+      background-color: rgb(128, 128, 128); /* same as var(--vscode-panelSectionHeader-background) but without transparancy */
+      text-align: left;
+      position: sticky; /* Lock the footer row to the bottom so it's always visible as rows are scrolled */
+      bottom: 0;        /* Don't forget this, required for the stickiness */
+    }
+
     #resultset th,
     #resultset td {
       padding: 5px 15px;
--- a/src/views/html.ts
+++ b/src/views/html.ts
@@ -21,7 +21,7 @@ export function getHeader(options: {withCollapsed?: boolean} = {}): string {
     }
 
     tfoot tr {
-      background-color: rgb(128, 128, 128); /* same as var(--vscode-panelSectionHeader-background) but without transparancy */
+      background-color: var(--vscode-multiDiffEditor-headerBackground);
       text-align: left;
       position: sticky; /* Lock the footer row to the bottom so it's always visible as rows are scrolled */
       bottom: 0;        /* Don't forget this, required for the stickiness */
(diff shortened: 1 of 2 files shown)


src/views/html.ts
```diff
@@ -20,6 +20,13 @@ export function getHeader(options: {withCollapsed?: boolean} = {}): string {
       top: 0;           /* Don't forget this, required for the stickiness */
     }
 
+    tfoot tr {
+      background-color: rgb(128, 128, 128); /* same as var(--vscode-panelSectionHeader-background) but without transparancy */
```
worksofliam, line 24: I don't think this is the right colour to use. I will share some screenshots, but it doesn't come through too well in Light Modern or Dark Modern like the header does. I also think we should use the CSS variables when we are able to instead of hardcoding colours.

https://code.visualstudio.com/api/references/theme-color

src/views/results/html.ts
```diff
@@ -85,21 +86,21 @@ export function generateScroller(basicSelect: string, isCL: boolean): string {
...
+                    appendRows(data.rows);
                   }
 
-                  const nextButton = document.getElementById("nextButton");
                   if (data.rows === undefined && totalRows === 0) {
-                    nextButton.innerText = 'Query executed with no result set returned. Rows affected: ' + data.update_count;
+                    document.getElementById(messageSpanId).innerText = 'Statement executed with no result set returned. Rows affected: ' + data.update_count;
                   } else {
-                    nextButton.innerText = noMoreRows ? ('Loaded ' + totalRows + '. End of data') : ('Loaded ' + totalRows + '. Fetching more...');
+                    document.getElementById(statusId).innerText = noMoreRows ? ('Loaded ' + totalRows + '. End of data') : ('Loaded ' + totalRows + '. More available...');
```
worksofliam, line 102: After trying it with `qsys2.columns`, it makes me debate the point of 'More available' going forward. Before you only saw it for a brief moment before it goes away again. I wonder if it should just say 'Loaded X' and 'Loaded X. End of data.'

worksofliam (changes requested):
Overall makes a lot of sense. I can see this new footer bar being used to also solve #102 down the line, since we have somewhere to put it now. See my comment about color and text. Other than that, it's decent.

image

image

src/views/html.ts
davecharron: Makes sense.  Did you have a specific CSS variable you'd suggest?

src/views/results/html.ts
davecharron: Not sure.  Without it there is no indication that there is more data, the user would just have to know that's the case because it _does not say **Done**_.

worksofliam: Perhaps it is just the ellipsis that is not needed. Just 'More available.'

src/views/html.ts
worksofliam: `--vscode-multiDiffEditor-headerBackground` seems like a good option so far.

image

image

image

worksofliam (approved):
This is a great change. A good setup for future features too

PR #220, opened 2024-03-26
Feature/result table column headings

### LINKED ISSUE #127: Ability to toggle file column labels on results view
Toggle switch between only column name and label. Not both

davecharron:
#### Overview
Add support for customizing result column headings.

- [x] New configuration setting for `Column headings`
  - [x] Options are `Name` and `Label`
        Note: `Label` requires the JDBC config 'Extended metadata' setting to be turned on.
- [x] New command for editing the setting from the results view
- [x] Listen for changes to the setting and apply to open results view
- [x] Bonus: Organized the Db2 for i preferences

#### Test
```sql
CREATE SCHEMA TESTHEAD;

CREATE OR REPLACE TABLE TESTHEAD.T1 (
            C1 INTEGER DEFAULT NULL,
            C2 VARCHAR(25) CCSID 37 DEFAULT NULL
        );

LABEL ON COLUMN TESTHEAD.T1 (
    C1 IS 'C1 Column Heading',
    C2 IS 'C2 Column Heading'
);

SELECT * FROM TESTHEAD.T1
```

Closes #127 Ability to toggle file column labels on results view

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,13 +51,8 @@
       {
         "id": "vscode-db2i",
         "title": "General",
-        "order": 1,
+        "order": 0,
         "properties": {
-          "vscode-db2i.autoRefreshSelfCodesView": {
-            "type": "boolean",
-            "description": "Enable auto-refresh for SELF Codes view when connecting to a system",
-            "default": false
-          },
           "vscode-db2i.pageSize": {
             "type": "number",
             "description": "Page size for Schema browser",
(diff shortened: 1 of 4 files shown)


worksofliam:
@davecharron Send help! What am I doing wrong to test this? Sorry for the big screenshot, but it shows:

* My select statement, with a `LABEL` statement to update the label of a column.
* I have 'Label' selected in my VS Code settings
* I have Extended metadata to true and saved

But can't get it to show up.

image

Through a little more digging, it looks like the server component isn't returning is still?

```json
"columns":[{"name":"EMPNO","type":"CHAR","display_size":6,"label":"EMPNO"},
```

davecharron:
> @davecharron Send help! What am I doing wrong to test this? Sorry for the big screenshot, but it shows:
You might need to reconnect after the LABEL ON.  We've seen issues where SQ caches metadata so it doesn't always show recent changes.

worksofliam:
Okay, I got it working. It did require a new job to get the changes. This worked:

```sql
select * from sample.employee;

LABEL ON COLUMN SAMPLE.employee (EMPNO IS 'Employee Number');
```

More comments coming.

worksofliam:
@davecharron So based on this PR, I think we should change one of the JDBCOptions default.

Perhaps, in the `SQLJob` constructor, where we pass [JDBCOptions](https://github.com/davecharron/vscode-db2i/blob/5819f096e2e5ddd39ad6e40d239e3259ab3bef9c/src/connection/sqlJob.ts#L79) as `options`, we should set `extended metadata` on by default for all jobs?

```ts
  constructor(public options: JDBCOptions = {}) {
    this.options["extended metadata"] = true;
  }
```

davecharron:
@worksofliam I'm not sure about changing the default.
The Javadoc for [JDBC Properties](https://www.ibm.com/docs/api/v1/content/ssw_ibm_i_75/rzahh/javadoc/com/ibm/as400/access/doc-files/JDBCProperties.html) indicates a potential performance hit when "extended metadata" is set to true.

> Setting this property to true may slow performance because it requires retrieving more information from the system.

worksofliam:
@davecharron Plus, users can store it in their custom config if they want it enabled always. Good point.

worksofliam (approved):

Issue #222, opened 2024-04-01
Db2 for i Error


hn000246:
![image](https://github.com/codefori/vscode-ibmi/assets/50391179/da6d4fbe-bd46-4499-8e8f-fbf68c680657)

Hello @worksofliam !!
I just upgrade db2 for i but i can see the Sql Job mánager anymore.
Do you know why? You can see the picture.

Best Regards,

worksofliam:
Please share the contents of your Code for IBM i output channel after it doesn't appear.

<img width="939" alt="image" src="https://user-images.githubusercontent.com/3708366/178739532-03c8ad30-84de-4698-9904-65938f2ef651.png">

worksofliam:
Please update to 0.10.0, which should solve your issue.

hn000246:
Thanks @worksofliam 
Best Regards.

PR #223, opened 2024-04-01
Cleanup of code that checks OS version and features

worksofliam:
* Cleans up the logic that determines the OS version, Db2 level and what features are enabled by moving all the logic into an IBM i class, which we can use to tie this kind of information to.
* Fixes a bug where when connecting to a system, but the user doesn't have correct authority to `group_ptf_info`, then it would 
    * a) not finish the startup process correctly, 
    * b) completely disabled SELF, and 
    * c) failed to start a job up automatically based on the configured option
* Fixes invalid hoover tooltip on SELF stack

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,8 +53,9 @@
         "title": "General",
         "order": 1,
         "properties": {
-          "vscode-db2i.autoRefreshSelfCodesView": {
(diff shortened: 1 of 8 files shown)


worksofliam:
This may fix #222

forstie (approved):
Tested on 7.4, looks great

PR #224, opened 2024-04-02
Add job id to result set footer

worksofliam:
Simply adds the executing job id to the footer of the result set.

image

### DIFF
--- a/src/views/results/html.ts
+++ b/src/views/results/html.ts
@@ -102,10 +102,6 @@ export function generateScroller(basicSelect: string, isCL: boolean): string {
                     appendRows(data.rows);
                   }
 
-                  if (data.jobId) {
-
-                  }
-
                   if (data.rows === undefined && totalRows === 0) {
                     document.getElementById(messageSpanId).innerText = 'Statement executed with no result set returned. Rows affected: ' + data.update_count;
                   } else {
--- a/src/views/results/html.ts
+++ b/src/views/results/html.ts
@@ -54,6 +54,7 @@ export function generateScroller(basicSelect: string, isCL: boolean): string {
           const basicSelect = ${JSON.stringify(basicSelect)};
           const htmlTableId = 'resultset';
           const statusId = 'status';
+          const jobId = 'jobId';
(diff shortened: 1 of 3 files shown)


davecharron:
I wonder if the job name text should be slightly less prominent than the results status.
Perhaps font that is grey and/or slightly smaller size.
The way it is displayed in the SQL Job Manager.
image

davecharron:
Another idea, perhaps requiring its own PR, would be to add a span to the footer, between the status and the job name, with the caption 'SQL'.  Hovering over it would display the formatted SQL that produced the results.

src/views/results/html.ts
```diff
@@ -96,10 +97,15 @@ export function generateScroller(basicSelect: string, isCL: boolean): string {
                     appendRows(data.rows);
                   }
 
+                  if (data.jobId) {
+
+                  }
```
davecharron, line 102: Seems there is nothing to do here, remove the block.

worksofliam:
@davecharron 

> Edited: Alternate idea would be to have the new span have the caption 'Details', that when hovered over, shows the job name and SQL. This approach de-clutters the footer, and makes adding details later more straightforward.

I like this idea a lot, but it should be a separate PR. Also, I don't know if from a web view we get the same cool hover support like we do with tree views.

src/views/results/resultSetPanelProvider.ts
```diff
@@ -41,9 +41,12 @@ export class ResultSetPanelProvider implements WebviewViewProvider {
 
           let queryResults = queryObject.getState() == QueryState.RUN_MORE_DATA_AVAILABLE ? await queryObject.fetchMore() : await queryObject.run();
 
+          const jobId = queryObject.getHostJob().id;
+
           data = queryResults.data;
```
davecharron, line 47: Doesn't look like we need `data`.  It's defined on line 29 and set here, but is never used.

davecharron (approved):

davecharron (approved):
Send it.

PR #227, opened 2024-04-17
Fix reference to non-existent variable

davecharron:
Fix a bug introduced by PR #224

A non-existent variable is referenced in the javascript in the generated HTML for displaying results.

Change to use the correct variable.

Fixes issue #228


worksofliam (approved):
Tested and works as expected. Good find!

PR #238, opened 2024-06-17
Enable result sets in webviews

worksofliam:
Adds a new button the execute dropdown to run the statement in a new view.

This will run the statement in a brand new tab instead of the standard webview.

### DIFF
--- a/package.json
+++ b/package.json
@@ -315,6 +315,12 @@
         "when": "editorLangId == sql && resourceExtname != .inb"
       },
       {
+        "command": "vscode-db2i.runEditorStatement.inView",
+        "key": "ctrl+alt+r",
+        "mac": "cmd+ctrl+r",
+        "when": "editorLangId == sql"
+      },
+      {
         "command": "vscode-db2i.editorExplain.withRun",
         "key": "ctrl+shift+r",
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -314,6 +314,12 @@
         "mac": "cmd+r",
         "when": "editorLangId == sql && resourceExtname != .inb"
       },
+      {
+        "command": "vscode-db2i.runEditorStatement.inView",
+        "key": "ctrl+alt+r",
+        "mac": "cmd+ctrl+r",
+        "when": "editorLangId == sql"
```
davecharron, line 321: Should this have the same resource extension check as at line 315?

davecharron (approved):

package.json
worksofliam: I didn't think so. It might be nice to take a statement from a Notebook and run it in a new Window

forstie:
Two comments:
1) For longer running SQL statements, I could not see the SQL cancel button when using the enhancement
2) It would be nice to give users a method to control the name of the tab/window that is being opened

forstie (approved):
I'm seeing the expected behavior now. Thanks

PR #240, opened 2024-06-19
activate vscode-data-table extension for notebooks

ajshedivy:
## render tables with vscode-data-table extension
- https://github.com/RandomFractals/vscode-data-table

image

Column stats:
image

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,9 @@
     "onLanguage:sql"
   ],
   "extensionDependencies": [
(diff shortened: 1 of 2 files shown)


ajshedivy:
There is a known issue with the json export functionality. In the Flat Data Grid view, there is an option to export data view as json or csv - current only csv works.

Issue: https://github.com/RandomFractals/vscode-data-table/issues/147

PR #242, opened 2024-06-20
integrate data preview extension

ajshedivy:
Addresses #240 

TODO:
- Add command to Run Dialog to open Data Preview
- review `preview` logic
- add data preview extension as dependency, or check if the extension exists
    - show message to user that `preview` requires data preview extension installed
- use `vscode.fs` instead of `fs/promises`

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,8 @@
     "onLanguage:sql"
   ],
   "extensionDependencies": [
-    "halcyontechltd.code-for-ibmi"
+    "halcyontechltd.code-for-ibmi",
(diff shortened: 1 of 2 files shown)


RandomFractals:
@ajshedivy great idea! Fair note: our data preview extension is very stale and out of date. 

Let me know how this PR pans out for you. We've been meaning to publish new Data Preview Pro Tools for some time now, but currently taking a break from unpaid VS Code extensions till MSFT makes it possible for devs to monetize them directly from VS Code marketplace.

PR #248, opened 2024-07-22
Get MTIs for a schema or table, improved index info

worksofliam:
* "Get Indexes" on tables now more detailed and includes MTI info where applicable
* Right click on either tables or schemas a new "Get MTIs" option will be available.

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -418,6 +418,11 @@
         "category": "Db2 for i"
       },
       {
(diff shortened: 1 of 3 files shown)


forstie (approved):
All 3 scenarios ( 2 new actions and 1 revised action) work well after the changes.

PR #252, opened 2024-07-31
Cleanup/filter_params

worksofliam:
As spotted by Julia, we were not using binding parameters in our filtering code. This was bad. Now we do!

Should have no impact in functionality. I tested the UX and the test cases are passing.


worksofliam:
Looks to be working as expected @julesyan - thank you!

image

image

PR #255, opened 2024-08-07
Ability to toggle self for specific job

worksofliam:
This PR adds the ability to toggle SELF to only show SELF logs for the selected job. 

When it is toggled on, we can also show the job log for the selected job.

image

image


ajshedivy:
added logic to only reset errors from active job in the filtered view

PR #258, opened 2024-08-08
SQL Variable view

worksofliam:
* Adds new view titled "SQL Variables"
* View can allow the user to see the value of SQL variables, data areas or some SQL statement
* Variables can be added by the user by clicking the variable button in the view header
* Variables can be removed by the user by using the minus button on each variable
* The view is automatically updated when a statement/notebook cell is run
* The view will make a suggestion (marked with a plus) to add a variable to the view after running a statement that returns one row and one column

#### Scene 1

1. `create or replace variable yourschema.helloworld varchar(10000) ccsid 1208 default 'scooby';`
2. navigate to `yourschema` in the Schema Browser and expand Variables
3. right click on your new `helloworld` variable and 'Add' it
4. expand the variable view and see the variable and value of it there
5. `set yourschema.helloworld = 'other value';` and see the value automatically change in the view

#### Scene 2

1. run `select count(*) from sometable` (make sure you can insert into this table) and see the result
2. expand the Variables view and see a new node with the plus (+) icon (wording includes 'suggestion' on the node)
3. click on the suggestion node and see the count variable node appear with the count value
4. run `insert into sometime (x, y, z)` and after execution see the count variable increase

### DIFF
--- a/package.json
+++ b/package.json
@@ -264,6 +264,13 @@
           "name": "SQL Error Logging Facility (SELF)",
           "when": "code-for-ibmi:connected && vscode-db2i:SELFSupported && vscode-db2i:jobManager == true",
           "visibility": "collapsed"
+        },
+        {
+          "type": "tree",
+          "id": "vscode-db2i.variables",
+          "name": "SQL Variables",
+          "when": "code-for-ibmi:connected && vscode-db2i:jobManager.hasJob",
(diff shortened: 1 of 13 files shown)


worksofliam:
Not sure now is the right time for a feature this large. Will come back to it in the future.

Issue #262, opened 2024-08-13
Schema Browser it doesn't show the materialized query tables (MQT) existing in my Lib


jdetrizep:
In the "DB2 for i" extension, the browser schema does not show the tables of materialized queries created in my library. It would be good to have the possibility of viewing the materialized query tables, I imagine that at the time of bringing the table catalog, they have a filter for the "Table Type" field and exclude the materialized tables, however, it would be very useful to be able to see the materialized tables that we have in our libraries.

PR #264, opened 2024-08-19
Also show MQTs in Tables node

worksofliam:
Implements #262.

Extends the Tables node to include MQTs.


forstie (approved):
Reviewed and tested. 
Works.

PR #273, opened 2024-09-04
Replace internal API with `mapepire-js`

worksofliam:
* Use brand new server component version 2.x, which is now known as Mapepire, instead of the 'server component'
* Inherit the SQLJob API from mapepire-js, but override the logic to use an SSH connection instead of a web socket connection with the server in 'single' mode


worksofliam:
All tests passing!

image

image

PR #277, opened 2024-09-17
Hover support

worksofliam:
Adds hover support for columns, procedures and functions.

* Rewrote caching mechanism
* Fetch SQL information when an SQL file is loaded


forstie (commented):
excellent updates after extensive testing

PR #279, opened 2024-09-21
Feature/formatter

worksofliam:
Adds the ability to format SQL documents.

#### Internal notes

Has major change to parser to include scalar functions in `getObjectReferences`. This might affect Source Orbit. It also fixes some issues with hover.

### To format a document:

1. Open SQL document
2. Write or add content
3. F1 -> 'Format Document'

### To configure SQL formatter

1. Open VS Code settingd (Command + Comma)
2. Navigate to Extension -> Db2 for i -> SQL Formatting (or search 'sql formatting')

### How we collect feedback

Reply with:

* Your SQL code that doesn't format correctly
* What you expect the code to look like
* The format settings you used (if not the default)


worksofliam:
Possible hover bug with `getObjectStatistics`?

```sql

select a.*, b.*
from qsys2.exit_program_info a, lateral (
  select *
  from table (
    qsys2.object_statistics(
      exit_program_library,
      '*pgm',
      exit_program
    )
  )
) b
where exit_point_name like 'qibm_qca_%_command%'
order by exit_point_name, exit_program_number;
```

PR #280, opened 2024-09-23
Check if remote server components actually needs an update

sebjulliand:
Related to this issue: https://github.com/codefori/vscode-ibmi/issues/2267

If `Config.getServerComponentName` fails to return the currently installed version of the server component (probably happens when installing the extension on a new VSCode instance since it's taken from the extension global storage?), the update manager will try to overwrite the existing file that is not writable (because `chmod 400` is run on it after it's uploaded).

To prevent the updater from failing, this PR adds a new check that will:
- Check if the remote component already exists
- If it exists, the local file's md5 hash is compared to the remote file's

If the md5 hashes are different, the remote file is made writable to ensure the upload will not fail.

This should prevent issues like https://github.com/codefori/vscode-ibmi/issues/2267 from happening.

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,7 +35,7 @@
     "lint": "eslint .",
     "pretest": "npm run lint",
     "language:test": "vitest",
-    "dsc": "mkdir -p dist && npx tsx src/dsc",
+    "dsc": "npx tsx src/dsc",
     "package": "vsce package",
     "vscode:prepublish": "rm -rf dist && npm run webpack && npm run dsc",
     "webpack": "vscd --clean && webpack --mode development",
--- a/src/connection/serverComponent.ts
+++ b/src/connection/serverComponent.ts
@@ -1,11 +1,15 @@
 import { getInstance } from "../base";
 
-import { Config } from "../config";
 import path from "path";
 import { OutputChannel, extensions, window } from "vscode";
+import { Config } from "../config";
 
 import { stat } from "fs/promises";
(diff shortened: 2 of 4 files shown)


sebjulliand:
Lovely, thanks!

worksofliam:
@sebjulliand I just tested this out. To be safe, I also deleted my `~/.vscode/*` contents (not the directory), but then the install on this PR fails because the directory exists, even though the .jar doesn't exist.

sebjulliand:
> @sebjulliand I just tested this out. To be safe, I also deleted my `~/.vscode/*` contents (not the directory), but then the install on this PR fails because the directory exists, even though the .jar doesn't exist.
> 
> When I checked back out to main, then it continued to install correctly.

How did it fail for you?
I deleted the folder's content before connecting (before testing the PR actually) and it went fine. Did you get an error or anything? 😓 

But I found an edge case: connecting, getting the install done, disconnecting, deleting the component, reconnecting: it does not reinstall the component. I'll fix that too.

sebjulliand:
@worksofliam turns out the `ServerComponent` `installed` flag was not reset after disconnecting - I fixed that!
I also made the `dsc` npm command OS agnostic while I was at it...it wasn't running fine on Windows 😛

worksofliam:
@sebjulliand Yup, that works now! Thanks for a lovely change.

worksofliam (approved):
Working as expected.

PR #290, opened 2024-10-14
Feature/continue2

ajshedivy:


index.d.ts
worksofliam, line 1: What is this interface file for? Is this the Continue interface? Is this not available on NPM somewhere instead of embedding it in our source?

package.json
```diff
@@ -558,6 +558,12 @@
         "category": "Db2 for i",
         "icon": "$(database)"
       },
+      {
+        "command": "vscode-db2i.jobManager.continue",
```
worksofliam, line 562: I think this needs an improved id. Perhaps `vscode-db2i.jobManager.focusContinue`

src/aiProviders/continue/continueContextProvider.ts
```diff
@@ -0,0 +1,173 @@
...
+  `;
+
+    try {
+      const result = await selected.job
+        .query<SelfCodeNode>(content)
+        .execute(10000);
+      if (result.success) {
+        const data: SelfCodeNode[] = result.data.map((row) => ({
+          ...row,
+          INITIAL_STACK: JSON.parse(row.INITIAL_STACK as unknown as string),
```
worksofliam, line 83: Be careful here. I've seen this `JSON.parse` fail because `row.INITIAL_STACK` is not valid JSON. I'd put this into a function: 

```
INITIAL_STACK: tryGetJson(row.INITIAL_STACK as unknown as string),
```

src/aiProviders/continue/continueContextProvider.ts
```diff
@@ -0,0 +1,173 @@
...
+    args: LoadSubmenuItemsArgs
+  ): Promise<ContextSubmenuItem[]> {
+    return [];
+  }
+}
+
+export async function registerContinueProvider() {
+  const provider = new db2ContextProvider();
+  const continueID = `Continue.continue`;
+  const continueEx = vscode.extensions.getExtension(continueID);
```
worksofliam, line 170: `continueEx` can be undefined if the extension is not installed.

Also, you need to check if it's active or activate it if not.

Like this: https://github.com/codefori/vscode-db2i/blob/main/src/extension.ts#L109

worksofliam: Also, if the extension is not installed, then all the new Continue specific commands need to be disabled.

worksofliam (changes requested):
Overall the code looks good, but we need to better handle when the Continue extension is not installed.  Your `registerContinueProvider` function needs to handle this I think. 

Based on whether Continue is installed or not, the Continue specific commands would need to be disabled. See how we're using `code-for-ibmi:connected` [here](https://github.com/codefori/vscode-db2i/blob/main/package.json#L277). Do a search for that string in the codebase, where you can see we set it to true/false based on the connection state.

index.d.ts
ajshedivy: Unfortunetly not (at least when we started this), is there a way to access an extensions index.d.ts without adding it as a dependency? https://docs.continue.dev/customize/tutorials/build-your-own-context-provider#extension-api-for-vscode

src/aiProviders/continue/continueContextProvider.ts
ajshedivy: if the parse fails, should we throw an error or just ignore that entry?

package.json
```diff
@@ -558,6 +558,12 @@
         "category": "Db2 for i",
         "icon": "$(database)"
       },
+      {
```
worksofliam, line 561: Add `enablement` to this command. 

`"enablement": "continueExtensionActive"`

package.json
```diff
@@ -676,6 +682,12 @@
         "category": "Db2 for i",
         "icon": "$(info)"
       },
+      {
```
worksofliam, line 685: Same thing with `enablement` here.

src/aiProviders/continue/continueContextProvider.ts
```diff
@@ -0,0 +1,192 @@
...
+  if (continueEx) {
+    if (!continueEx.isActive) {
+      await continueEx.activate();
+    }
+  
+    isContinueActive = true;
+    const continueAPI = continueEx?.exports;
+    continueAPI?.registerCustomContextProvider(provider);
+    vscode.commands.executeCommand('setContext', 'continueExtensionActive', true);
```
worksofliam, line 189: We should follow the standard we have for setting the context. Usually our context names are closely unique to this extension. Perhaps this instead?

```ts
commands.executeCommand(`setContext`, `vscode-db2i:continueExtensionActive`, isContinueActive);
```

src/views/jobManager/selfCodes/contributes.json
```diff
@@ -92,6 +92,12 @@
         "category": "Db2 for i",
         "icon": "$(info)"
       },
+      {
```
worksofliam, line 95: Add enablement here also.

src/views/jobManager/selfCodes/contributes.json
```diff
@@ -147,6 +153,11 @@
           "command": "vscode-db2i.self.displayDetails",
           "when": "view == vscode-db2i.self.nodes && viewItem == selfCodeNode",
           "group": "navigation"
+        },
+        {
```
worksofliam, line 157: Enablement here also.

worksofliam (changes requested):
Mostly simple things.

Setting `enablement` correctly in the new commands and one note about the context name.

worksofliam (approved):
Reviewed together and tested together. Looks excellent!

PR #294, opened 2024-11-04
Fix/chat history

ajshedivy:


src/connection/sqlJob.ts
```diff
@@ -159,6 +159,25 @@ export class OldSQLJob extends SQLJob {
...
+    if (schema) {
+      const upperSchema = schema.toUpperCase();
+      const result = await this.execute(`set current schema = ?`, {parameters: [upperSchema]});
+      if (result.success) {
+        const config = getInstance().getConfig();
+        const currentLibrary = config.currentLibrary.toUpperCase();
+    
+        if (upperSchema !== currentLibrary) {
+          config.currentLibrary = upperSchema;
+          await getInstance().setConfig(config);
```
worksofliam, line 172: You actually don't need to update the Code for IBM i config, just the selected job options.

`this.options.libraries[0] = ``;`

The selected job and the connection config have two separate current libraries.

src/connection/sqlJob.ts
```diff
@@ -159,6 +159,25 @@ export class OldSQLJob extends SQLJob {
     return connectResult;
   }
 
+  async setCurrentSchema(schema: string): Promise<QueryResult<any>> {
+    if (schema) {
+      const upperSchema = schema.toUpperCase();
```
worksofliam, line 164: Instead of `toUpperCase`, use this API we have:

```ts
Statement.delimName(schema, true)
```

worksofliam (changes requested):
Two simple changes!

PR #297, opened 2024-11-06
ignore context item it its empty

ajshedivy:


worksofliam (approved):
Nice additional check there. I will merge and release.

PR #301, opened 2024-11-24
Add "View authorities" option on SQL objects. Resolves codefori/vscode-db2i#266

dferrand:
Add "View authorities" option on SQL objects.


src/views/schemaBrowser/contributes.json
```diff
@@ -95,6 +95,11 @@
         "title": "Set filter",
         "category": "Db2 for i",
         "icon": "$(search)"
+      },
+      {
+        "command": "vscode-db2i.viewAuthorities",
+        "title": "View authorities",
```
forstie, line 101: Seems like "View authorities" should be: "View Permissions"

forstie (approved):
One question as to the best name for the Action.

Other than that, looks good. I look forward to trying it out.

src/views/schemaBrowser/contributes.json
dferrand: I changed "View authorities" to "View permissions"

worksofliam:
Tested and working.

PR #307, opened 2024-12-17
Implement syntax checking features

worksofliam:
Introduce a problem provider that highlights SQL syntax issues and allows for configurable syntax check intervals. This feature will only syntax check as you are typing. If you're working in a document made up of multiple statements, it will only check the statement you're currently working on while you type. You can use the newly added button to check an entire document.

How to test while editing:

1. Connect to IBM i with Code for IBM i
2. Open an SQL document (new or existing)
3. Start typing. There is wait period before the syntax is checked (configurable), but the default is 1.5 seconds.

Test the _Check Syntax_ button:

1. Connect to IBM i with Code for IBM i
2. Open an SQL document (new or existing)
3. Press the _Check Syntax_ button, which appears next to the Run button in the editor bar in the top right.

image


p-behr:
It sure appears to provide the same syntax check as running the SQL in ACS Run SQL Scripts :-)

chrjorgensen (changes requested):
@worksofliam Looking good and working fine with most samples. 👍 

I found one issue:

- connect to IBM i (7.5 in my test)
- open a new SQL document
- enter the following source:

```sql
update qtemp.temp_rcpout
   set COL1 = '123'
     , COL2 = replace( COL2, 'a', 'A' )
     , COL3 = replace( COL3, 'b', 'B' )
     , COL4 = 'L'
     , COL5 = '/dir1/'
     , COL6 = 0
```
- press the button to syntax check

I then get this:

billede

worksofliam:
@chrjorgensen Have another look. I added some additional logic to handle when the message ID is not returned, which is why you were seeing the blank message. It actually looks like this could be a bug in the system API, and IMO shouldn't be impactful to this PR.

I also made it so `01` SQL states are warnings and not errors.

worksofliam:
@chrjorgensen I added a new VS Code configuration option to determine if warnings should be shown. It is false by default.

julesyan:
There seems to be some issue with the syntax checking running when I comment out a line. I have the following
```
stop;
stop;
stop;
```
Which produces three errors. If i press CMD + / that line gets checked correctly. But if i comment out a second line, that second line does not get checked and the error remains. Same if i comment out a third line. The order of the lines do not seem to matter.

julesyan:
Another issue found for CL commands. If i start with `CL` on a line it syntax check and gives an error. Then when i continue writing the rest of the line where it is `CL: <command>;` it never rechecks to remove the error. 
Same happens if i delete the `:` on an existing CL command, wait for syntax check, then add it back in.

forstie (approved):
Good job dealing with the side issues found during testing.
A very nice addition indeed.

julesyan (approved):

julesyan (commented):
> There seems to be some issue with the syntax checking running when I comment out a line. 
Actually this is still happening

PR #313, opened 2025-01-05
Right click option on SQL objects to view object locks

janfh:
A new context option to view object locks similar to the one found in Schemas in ACS

image

image


worksofliam (approved):
Tested and working as expected. Thanks!

PR #316, opened 2025-01-08
Prevent sql tokenizer from assuming FOR is a statementType when used with data-type

### LINKED ISSUE #315: Statement with data-type ... FOR BIT DATA fails
It looks like the SQL tokenizer thinks that **FOR** is always a statementType, while it can be used with data-type as well, e.g. "... VARCHAR(100) FOR BIT DATA ..."

To reproduce, run the following statement:
```
select 
  cast(x'01' as char(1) for bit data) as something,
  case when 1=1 then 'makes sense' else 'what?' end as something_else
  from sysibm.sysdummy1;
```
Result:
image

Debugging the tokenizer shows that "**for**" is given the statementType:

image

janfh:
Fixes #315 

Modified fixStatement to not assign statementType for "**FOR**" when:
* Previous token is "closebracket"
* Token after the next is the word "**DATA**"

### DIFF
--- a/src/language/sql/tests/statements.test.ts
+++ b/src/language/sql/tests/statements.test.ts
@@ -149,6 +149,17 @@ parserScenarios(`Object references`, ({newDoc}) => {
     expect(obj.alias).toBe(`a`)
   });
 
+  test('SELECT: for in data-type (issue #315)', () => {
+    const document = newDoc([
+      `select cast(x'01' as char(1) for bit data) as something,`,
+      `case when 1=1 then 'makes sense' else 'what?' end as something_else`,
+      `from sysibm.sysdummy1;`
+    ].join(`\n`));
+  
+    expect(document.statements.length).toBe(1);
+    expect(document.statements[0].type).toBe(StatementType.Select);
+  });
+
   test('SELECT: Simple qualified object with alias (no AS)', () => {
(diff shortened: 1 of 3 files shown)


worksofliam:
@janfh This is a great change. But, since you are making a change in a fairly critical file, it'd be nice to have a test case for this change. Might I recommend you add a test for this in `src/language/sql/tests/tokens.test.ts`? I would usually use the statement that you found this issue with:

```
select 
  cast(x'01' as char(1) for bit data) as something,
  case when 1=1 then 'makes sense' else 'what?' end as something_else
  from sysibm.sysdummy1;
```

If you need any help - let me know. I am very grateful for this PR.

janfh:
> @janfh This is a great change. But, since you are making a change in a fairly critical file, it'd be nice to have a test case for this change. 

@worksofliam : I've added a test case. But please let me know if you would like something else, or feel free to adjust any of these changes :)

worksofliam:
@janfh Looking good. I added one more additional test in a commit of my own to ensure the statement length (as this is what your code fixed). I can now merge. Thanks again!

worksofliam (approved):

worksofliam:
@janfh I should mention that this should come out soon. I am trying to get some edge cases of the syntax checker done, and hopefully we can release this fix along with that.

PR #317, opened 2025-01-08
Assertions in tokens.test.ts

janfh:
I noticed that some tests were missing e.g. "toBe" assertion on expect, and would always pass. Example:

image

What do you think @worksofliam ?


worksofliam:
@janfh Yup, this is a welcome change! I am going to hold off and wait for the syntax checker PR to merge before this one. That PR has an upgrade to vitest which will re-enable debugging of tests inside of VS Code.

worksofliam:
@janfh Can you please bring in commits from main before I merge this? Thanks!

janfh:
@worksofliam : Merge or close? :)

worksofliam (approved):
Tested and working

PR #326, opened 2025-01-31
Fix linebreaks in generated SQL

chrjorgensen:
The Generate SQL output use the result set from the `QSYS2.GENERATE_SQL` procedure, but this is using a source member (in QTEMP) as intermediate storage and thus have a fixed length source line - 80 characters by default.

This is a problem, since identifiers can be split by a linebreak.

An example is the generated source for view `QSYS2.OBJECT_LOCK_INFO` where the lines containing the SELECT is broken in the middle of identifiers and functions:

billede

This can not be fixed by the SQL formatter, which will replace a linebreak in an identifier by a space:

billede

This PR will change the use of the `QSYS2.GENERATE_SQL` procedure to generate a streamfile instead of a source member and read this file. This will eliminate the problem with linebreaks in identifiers:

billede

After SQL formatting the source will look as expected:

billede

### DIFF
--- a/src/database/schemas.ts
+++ b/src/database/schemas.ts
@@ -1,6 +1,6 @@
 
+import path from "path";
 import { getInstance } from "../base";
-
 import { JobManager } from "../config";
 
 export type SQLType = "schemas" | "tables" | "views" | "aliases" | "constraints" | "functions" | "variables" | "indexes" | "procedures" | "sequences" | "packages" | "triggers" | "types" | "logicals";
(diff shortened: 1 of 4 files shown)


src/database/schemas.ts
```diff
@@ -220,11 +220,21 @@ export default class Schemas {
...
+                              , CREATE_OR_REPLACE_OPTION => '1', PRIVILEGES_OPTION => '0'
+                              , DATABASE_SOURCE_FILE_NAME => '*STMF'
+                              , STATEMENT_FORMATTING_OPTION => '0'
+                              , SOURCE_STREAM_FILE => '/tmp/Q_GENSQL_' concat current_user concat '.sql'
+                              , SOURCE_STREAM_FILE_END_OF_LINE => 'LF'
+                              , SOURCE_STREAM_FILE_CCSID => 1208 )`
     ].join(` `), { parameters: [object, schema, internalType] });
+    const lines = await JobManager.runSQL<{ LINE: string }>(
+      `select LINE
+         from table( QSYS2.IFS_READ( PATH_NAME => '/tmp/Q_GENSQL_' concat current_user concat '.sql' ) )`
```
worksofliam, line 234: At this point, should we just use our own Code for IBM i API to get the contents of the file?

chrjorgensen: We could... and we could also use our own C4i API to run the SQL and get the result in one step instead of these multiple calls. And maybe use the temporary streamfile API?

I'm quite busy currently with bigger projects at work, so we either 
1. wait for me to get some spare time
2. let you code it or
3. put a TODO in the source.

Which do you prefer? 😉

worksofliam: I am happy to do it actually! I will have a go at it on Monday morning. Using the temp streamfile is also a good idea.

src/database/schemas.ts
```diff
@@ -220,23 +220,27 @@ export default class Schemas {
...
-    );
-
-    const generatedStatement = lines.map( elem => elem.LINE ).join(`\n`);
-
-    return generatedStatement;
+    const instance = getInstance();
+    const connection = instance.getConnection();
+
+    const result = await connection.withTempDirectory<string>(async (tempDir) => {
+      const tempFilePath = path.posix.join(tempDir, `generatedSql.sql`);
```
chrjorgensen, line 227: @worksofliam Don't we have an API for generating a unique filename? Your code will collide between users, if both have `/tmp` as temp directory.

worksofliam: @chrjorgensen I believe the API you're referring to generates a unique name for a file you're going to use over and over again. But since this API, `withTempDirectory`, deletes the directory after, it is more suited here.

chrjorgensen: @worksofliam Ah, didn't know the `withTempDirectory` function... I should, though, since it was created in April last year by @sebjulliand ...! 😊

julesyan (approved):

PR #352, opened 2025-03-15
Ability to run multiple statements

worksofliam:
This adds new options under the play button:

Shown when there is no selection:

* Run all statements in the editor
* Run all statements from cursor

Shown when there is a selection:

* Run selected statements

This will execute all statements, but only show the result set for the last statement. Prefixes like `cl`, `csv`, etc, are still allowed and will work as they do when running single statements.

If a statement is execute and it errors, then no more statements will be executed.

If you use `stop`, then the statement execution will stop.

### DIFF
--- a/package.json
+++ b/package.json
@@ -437,6 +437,21 @@
         "icon": "$(window)"
       },
       {
+        "command": "vscode-db2i.runEditorStatement.multiple.all",
+        "title": "Run all statements",
+        "category": "Db2 for i"
+      },
+      {
+        "command": "vscode-db2i.runEditorStatement.multiple.selected",
+        "title": "Run selected statements",
+        "category": "Db2 for i"
+      },
+      {
+        "command": "vscode-db2i.runEditorStatement.multiple.from",
+        "title": "Run statements from cursor",
+        "category": "Db2 for i"
+      },
+      {
         "command": "vscode-db2i.statement.cancel",
         "title": "Cancel",
         "category": "Db2 for i",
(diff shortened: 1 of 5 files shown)


julesyan (changes requested):
Couple of issues and questions I have run into so far: 

- For run selected: 
    - does not work for partial selection, i get "No statements to run". I expect it to automatically pick up the two statements because there are two statements highlighted, regardless if it is fully highlighted
        image
    - similar, if i highlight a full statement but partially highlight another statement, it will only run the fully highlighted statement. Similar to above I expect it to run both
        image
- General:
    -  when an error happens in one of the statements it produces both a pop up error and an error in the IBM i view. Not sure if that was intended, but a bit annoying as a user. If I already have the error in the window, I do not need a notification as well. Can discuss furthur if you disagree
    - Why cant run all and run from cursor also be options when things are selected? 
    - Not sure if this is something you want to implement, but since we have the syntax checking, if there are any errors, maybe we shouldnt be allowed to run all? The only caveat being if its actual valid syntax and the checker is incorrect. If so then maybe we can give a warning that "the syntax is incorrect, would you like to still run all". 
    -  if a statement is taking awhile, and i happen to edit the file while i wait, will that break the upcoming statement? Or are we saving the statements before we start the whole run? 

In better news, I like that it highlights which statement is currently running!

julesyan:
Adding on, it does seem like you are saving the statements, not sure how that works when a file is edited though (is the get of the document a snapshot?). 
Is there a reason we cannot run CL commands in the multiple?

worksofliam:
@julesyan 

* For run selected, it will now correctly select the statements without the entire statement needing to be encapsulated in the selection range.
* The popup error is gone. Didn't realize double errors were appearing during my testing.
* _Run all_ is now always shown, but _Run from selected_ is still only available when no selection is made
* `cl:` now correctly throws an error when it doesn't run correctly.
* No plans to do anything with syntax checking enablement. You can still run single statements even if the syntax is invalid.

If you edit a document while it is running the statements, they will run as the statements were run it was initiated, but the ranges may be incorrect when it comes around the the next statement being executed is selected. We could add some logic to re-parse the document after every statement is run to grab the current changes.. though that might be slow for very very large documents.

forstie:
1) When run from or run all are in motion, can we have the "Cancel SQL" thing permitted?
2) Do you think some users will want key short-cuts for the new RUN actions?
3) When you code a stop; after a query, VS incorrectly says that the query is "still running".
image

--
-- Watch compare progress / activity  
--
SELECT *
   FROM TABLE (
         qsys2.active_job_info(
            job_name_filter => 'IFSCOMP', subsystem_list_filter => 'QBATCH')
      );
stop;

worksofliam:
@forstie New build is coming.

1. Fixed
2. Not sure yet...
3. Also fixed 😄

julesyan:
> If you edit a document while it is running the statements, they will run as the statements were run it was initiated, but the ranges may be incorrect when it comes around the the next statement being executed is selected. We could add some logic to re-parse the document after every statement is run to grab the current changes.. though that might be slow for very very large documents.

I think if you just pre-parse all the statements out, then run it all, that should avoid accidentally messing up the run. I dont think we should reparse the document and pick up new changes, that seems unnecessary overhead. Essentially just run it as a snapshot of when the button was hit

julesyan:
Aside from the issue with CL vs cl, looks good!

worksofliam:
@julesyan A new commit is in to make the prefixes case insensitive.

julesyan (approved):

forstie:
Everything is looking super now. Thanks for the adjustments!

forstie (commented):
Excellent.

forstie (approved):

forstie:
Run Selected works correctly after the change, thank you

forstie (approved):
Looks good.
The only thing to consider is short cuts to these great new actions.

PR #353, opened 2025-03-16
Graph based Visual Explain

worksofliam:
Introduce a new `CytoscapeGraph` class to manage graph elements and display them in a webview. Enable node selection to show detailed information.

Tasks:

* [ ] Merge in missing commits from `main`
* [ ] Fix that 'Run and explain' doesn't open the view, but 'Explain without running does'
* [ ] Improve formatting of nodes in webview
* [ ] **stretch**: Determine if coloring each node can be based on the old logic
   * https://github.com/codefori/vscode-db2i/blob/main/src/views/results/explain/doveTreeDecorationProvider.ts#L38 
* [ ] **stretch**: See about embedding icons onto nodes
   * Here are the attached icons for each different node type: https://github.com/codefori/vscode-db2i/blob/main/src/views/results/explain/doveResultsView.ts#L13 
   * Perhaps you need this to embed the icons into the webview? https://github.com/microsoft/vscode-codicons
* [ ] Attach description to each node if applicable
   * Old logic here: https://github.com/codefori/vscode-db2i/blob/main/src/views/results/explain/doveResultsView.ts#L126 
* [ ] Stop using remote URL to pull in cytoscape and embed into the build


SanjulaGanepola:
Closing this PR as new visual explain graph PR is being worked on

PR #357, opened 2025-03-18
Display time to execute when running SQL statements

RobertNash1:
https://github.com/codefori/vscode-db2i/issues/298

Resolved the ticket - added in performance metric (in milliseconds)

### DIFF
--- a/src/views/results/html.ts
+++ b/src/views/results/html.ts
@@ -379,7 +379,7 @@ export function generateScroller(basicSelect: string, isCL: boolean, withCancel?
                   if (data.rows === undefined && totalRows === 0) {
                     document.getElementById(messageSpanId).innerText = 'Statement executed with no result set returned. Rows affected: ' + data.update_count;
                   } else {
-                    document.getElementById(statusId).innerText = (noMoreRows ? ('Loaded ' + totalRows + '. End of data.') : ('Loaded ' + totalRows + '. More available.')) + ' ' + (updateTable ? 'Updatable.' : '');
+                    document.getElementById(statusId).innerText = (noMoreRows ? ('Loaded ' + totalRows + ' rows in ' + data.executionTime + 'ms. End of data.') : ('Loaded ' + totalRows + ' rows in ' + data.executionTime + 'ms. More available.')) + ' ' + (updateTable ? 'Updatable.' : '');
                     document.getElementById(jobId).innerText = data.jobId ? data.jobId : '';
                     document.getElementById(messageSpanId).style.visibility = "hidden";
                   }
--- a/src/views/results/resultSetPanelProvider.ts
+++ b/src/views/results/resultSetPanelProvider.ts
@@ -99,9 +99,9 @@ export class ResultSetPanelProvider implements WebviewViewProvider {
 
               if (this.currentQuery.getState() !== "RUN_DONE") {
                 setCancelButtonVisibility(true);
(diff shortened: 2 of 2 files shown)


src/views/results/resultSetPanelProvider.ts
```diff
@@ -112,7 +112,8 @@ export class ResultSetPanelProvider implements WebviewViewProvider {
                   columnHeadings: Configuration.get(`resultsets.columnHeadings`) || 'Name',
                   queryId: this.currentQuery.getId(),
                   update_count: queryResults.update_count,
-                  isDone: queryResults.is_done
+                  isDone: queryResults.is_done,
+                  executionTime: (endTime - startTime).toFixed()
```
worksofliam, line 116: So this is being sent for every request (e.g. the first request and the following page requests).

We actually only care about the execution time of the `execute` call and not all of them.

You will have to add some code to only send the execution time for the first execute (perhaps you base it on `getState()`), and `undefined` for the following requests. The frontend will need to be updated to account for this.

worksofliam (changes requested):

janfh:
FYI: Execution time is added to the server response from mapepire. Probably it doesn't make any difference unless there is some network latency between vscode and the server. Execution time will be available when mapepire is bumped.

https://github.com/Mapepire-IBMi/mapepire-js/pull/50

src/views/results/html.ts
```diff
@@ -379,7 +379,12 @@ export function generateScroller(basicSelect: string, isCL: boolean, withCancel?
                   if (data.rows === undefined && totalRows === 0) {
                     document.getElementById(messageSpanId).innerText = 'Statement executed with no result set returned. Rows affected: ' + data.update_count;
                   } else {
-                    document.getElementById(statusId).innerText = (noMoreRows ? ('Loaded ' + totalRows + '. End of data.') : ('Loaded ' + totalRows + '. More available.')) + ' ' + (updateTable ? 'Updatable.' : '');
+                    if (data.executionTime > 0) {
```
worksofliam, line 382: On the backend you're using `toFixed()` which returns a string, but here you're doing a numerical comparison. IMO, your backend code could send `undefined` on the second time around.

```ts
let executionTime: number|undefined;
if (this.currentQuery.getState() == "RUN_MORE_DATA_AVAILABLE") {
  queryResults = await this.currentQuery.fetchMore();
}
else {
  const startTime = performance.now();
  queryResults = await this.currentQuery.execute();
  const endTime = performance.now();
  executionTime = (endTime - startTime)
}

this._view.webview.postMessage({
  // ...
  executionTime //will send either a number or undefined
});
```

Then your frontend can do the `toFixed` calculation.

```js
if (data.executionTime) {
  document.getElementById(statusId).innerText = (noMoreRows ? ('Loaded ' + totalRows + ' rows in ' + data.executionTime.toFixed() + 'ms. End of data.') : ('Loaded ' + totalRows + ' rows in ' + data.executionTime.toFixed() + 'ms. More available.')) + ' ' + (updateTable ? 'Updatable.' : '');
}
```

worksofliam (changes requested):
This PR is really close! Just a small snag on some types and we should be good to go.

worksofliam (approved):
LGTM. Thanks @RobertNash1

worksofliam (approved):

PR #368, opened 2025-04-11
Generate RPG data structure from result metadata

janfh:
Feature to generate RPG data structure from column metadata.

Added statement qualifier "rpg". E.g:

````
rpg: select cast('' as varchar(100)) as my_text, current_timestamp as my_timestamp, cast(1.0 as numeric(13, 2)) as my_val from sysibm.sysdummy1;
````
generates:
````
**free

-- Row data structure
dcl-ds row_t qualified template;
  my_text varchar(100);
  my_timestamp timestamp;
  my_val zoned(13 : 2);
end-ds;
````

### DIFF
--- a/src/views/results/index.ts
+++ b/src/views/results/index.ts
@@ -18,7 +18,7 @@ import { updateStatusBar } from "../jobManager/statusBar";
 import { DbCache } from "../../language/providers/logic/cache";
 import { ExplainType } from "../../connection/types";
 
-export type StatementQualifier = "statement" | "update" | "explain" | "onlyexplain" | "json" | "csv" | "cl" | "sql";
+export type StatementQualifier = "statement" | "update" | "explain" | "onlyexplain" | "json" | "csv" | "cl" | "sql" | "rpg";
 
 export interface StatementInfo {
   content: string,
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/views/results/index.ts [hunk 1]
```diff
@@ -375,7 +375,69 @@ async function runHandler(options?: StatementInfo) {
...
+          if (statementDetail.statement.type !== StatementType.Select) {
+            vscode.window.showErrorMessage('RPG qualifier only supported for select statements');
+          } else {
+            chosenView.setLoadingText(`Executing SQL statement...`, false);
+
+            setCancelButtonVisibility(true);
+            updateStatusBar({executing: true});
+            const result = await JobManager.runSQLVerbose(statementDetail.content, undefined, 1);
+            setCancelButtonVisibility(false);
+            let content = `**free\n\n`
```
worksofliam, line 389: This logic needs to be in a unique function where the `metadata.columns` are the parameter.

worksofliam (changes requested):
I left a comment about moving logic to another function.

I don't want the logic to generate the content in this function really. We should move it elsewhere so we can write test cases for it.

Please make this change and then lmk when to look again.

src/views/results/index.ts
```diff
@@ -480,6 +493,53 @@ async function runHandler(options?: StatementInfo) {
   }
 }
 
+async function statementToRpgDs(statement: ParsedStatementInfo) : Promise<string> {
+  setCancelButtonVisibility(true);
+  updateStatusBar({executing: true});
+  const result = await JobManager.runSQLVerbose(statement.content, undefined, 1);
+  setCancelButtonVisibility(false);
```
worksofliam, line 500: Sorry to be picky. I don't think this should be in this function. This function should purely generate code and nothing else, and therefore the parameter to the function should be the result of `runSQLVerbose`.

worksofliam (changes requested):
Left another review.

janfh:
@worksofliam : Where do you want the tests? A new testsuite? Please continue to comment if you want me to change something else :) I will continue working on this tomorrow

worksofliam:
@janfh For my own sake of mind, I have had to move this to fix day 2 (Wednesday 23rd) and this will be the first PR I look at then. I may get to it over the weekend if I find a moment. Thanks for your work here!

Also, let's chat on chat.ibmioss.org sometime.

worksofliam:
@janfh Is there any reason you aren't using this `switch` statement on line 409 to check the `.qualifier` instead of adding a new `else if` block?

```ts
switch (statementDetail.qualifier) {
```

worksofliam:
@janfh i guess `runSqlVerbose` can always be used, but i see your point. Leave it as it is for now!

Going to do some final tests in the morning on this one and it should be good to go.

worksofliam (approved):

PR #370, opened 2025-04-16
Show column metadata as header row tooltip in result view

janfh:
Added column metadata to title property of header-cell in the result view. It will show the column data type, precision and scale.

image

How to test:
Hover over column headers to view data type


worksofliam (approved):
Seriously love this! A simple and effective change.

PR #381, opened 2025-04-20
Use date format = iso

### LINKED ISSUE #251: SQL gives different result set between VS code / ACS / greenscren
hi,

when trying only to use VS-Code i stumbled on something strange.
when looking at 1 date field the result set in VS code shows null, but actually it is *loval date.
even when i have a where 'Is not null' it still gives the *loval fields in the result set.

i think VS code makes the *loval value to a null in the result set but the query does not translate Null to *loval
353338660-e2b11fab-a8ba-4fe7-b20c-b6eae04bd52c

if needed more info you can always ask.

Ps. working on the latest releases of the plugins/ vs code

janfh:
Fixes #251 

Modified manager.ts#newJob to use date format = iso (jdbc props). I believe this is the default for ACS RSS.

More info:
https://www.ibm.com/support/pages/how-does-toolbox-jdbc-driver-deal-dates-1940-or-after-2039

### DIFF
--- a/src/connection/manager.ts
+++ b/src/connection/manager.ts
@@ -35,7 +35,8 @@ export class SQLJobManager {
         "full open": false,
         "transaction isolation": "none",
         "query optimize goal": "1",
-        "block size": "512"
+        "block size": "512",
+        "date format": "iso"
       }));
 
       try {


worksofliam (approved):
Looking good!

PR #382, opened 2025-04-21
Hide schema browser actions from command palette

janfh:
Hide schema browser actions that doesn't make sense to show in the command palette


worksofliam (approved):
A welcome PR for declutting that command pallete. Thank you!

Issue #256, opened 2024-08-07
Bug Report: IBM i Results Panel Doesn't Display Embedded Spaces


Wright4i:
## Description

When the results are shown in the `Results` panel any embedded spaces in a field are squashed into a single space.

## To Reproduce

```sql
SELECT 
    firstnme CONCAT ' ' CONCAT midinit CONCAT ' ' CONCAT  lastname AS fullname,
    firstnme CONCAT ' ' CONCAT midinit CONCAT '      ' CONCAT lastname AS extraspaces
FROM 
    sample.employee
```

## Expected behavior
Show every space.
![image](https://github.com/user-attachments/assets/a2b00ed5-f1c8-4c77-a808-881500bc1191)

## Screenshots
![image](https://github.com/user-attachments/assets/e1a3b9fa-c2ec-4517-97ec-0e06ee6af33c)

## Additional Information
The Trace shows the JSON had the spaces returned so the issue likely lies outside of the data fetching mechanism.

```html
<hr>
<b>[DATASTREAM_OUT]: </b><i>2024-08-07.13.19.58.918</i>
<font color="DarkSlateBlue">
<blockquote>
<pre>
[...]

Issue #302, opened 2024-12-05
Select statement not showing whitespace


Max-crosskey:
**Describe the bug**

Select statement doesn't include leading blanks in character field, for example if the field has value '        X' then the select and generated insert statement becomes 'X'-

**To Reproduce**

Insert leading blanks into a character field, then do select from DB2 for i Extension.

**Expected behavior**

Would expect the return statement to include leading blanks and be '        X'

**Screenshots**

VSC:
select
<img width="49" alt="image" src="https://github.com/user-attachments/assets/2a18a24a-5e29-46ef-acbf-48939171c6ae">
insert
<img width="70" alt="image" src="https://github.com/user-attachments/assets/0df07cfd-9ca1-4126-b65f-d95d340783ed">

STRSQL:
<img width="104" alt="image" src="https://github.com/user-attachments/assets/e81adddb-8d97-4f93-83dd-43bbe22aa334">

**Environment**

 - OS:  Windows 10
 - Extension Version 1.7.0

**Other**

Thought it might be a setting for the job but couldn't find it so apologies if I'm just blind

worksofliam:
Similar to #256.

@ThePrez is there something in the Mapepire that is trimming spaces from string fields?

dferrand:
I think there are two reasons why we don't see those whitespaces:

1/ It looks like Mapepire does trim white spaces, I suspect it is the <pre>.trim()</pre> on the <pre>cellDataForResponse = cellData.toString().trim();</pre> line in BlockRetrievableRequest.java.
Trimming trailing whitespaces might make sense for fixed length fields since they convey no information. I think they should be kept for varying length fields (even if SQL ignores them for comparisons, they are technically different values).
Leading whitespaces should never be trimmed.

2/ Even if the whitespaces were not trimmed by Mapepire, the result is displayed in a div in a td where the default html behavior will be to trim leading whitespaces (and collapse multiple whitespaces for that matter). This can be solved here with css (white-space: pre;).

worksofliam:
Solved with monospace

PR #384, opened 2025-04-22
Don't collapse whitespaces in results and output null in italic

dferrand:
Solves #302.

mapepire-server before 2.1.8 will trim leading whitespaces.


dferrand:
Also fixes #256

worksofliam (approved):
Nice. Thanks!

image

PR #391, opened 2025-05-02
RPG symbol name using name or label. Remove non-alphanum chars

### LINKED ISSUE #390: Generate RPG data structure uses label instead of fieldname
the new rpg: function uses labels for field names in the generated data structure instead of the actual field names. 

because we use the labels for descriptions the generated DS looks like this: 

`dcl-ds row_t qualified template;`
   `änderungs-          benutzer              char(10);`
`end-ds;`

Is there a setting to change this behavior ?

janfh:
Fixes #390

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "vscode-db2i",
-  "version": "1.11.0",
+  "version": "1.12.0",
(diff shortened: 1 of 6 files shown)


janfh:
> @janfh Is this intentionally a draft PR?

Please have a look and let me know what you think :)

PR #355, opened 2025-03-17
Bump Mapepire version to 2.2.2

worksofliam:
Update the Mapepire version to 2.2.1 in the codebase.

Run: `npm run dsc` before debugging.

* Updates Mapepire-server
* Updates Mapepire-client
* Cleanup how objects are fetched significantly

PR #398, opened 2025-06-03
sql-prefix to take generated columns into account

### LINKED ISSUE #392: `sql:` prefix to take generated columns into account
The `sql:` support isn't skipping generated columns, so the SQL cannot be used as generated.

```sql
CREATE TABLE THEBANK.CUSTOMERS ( 
	CUSTOMER_ID FOR COLUMN CUSTID     INTEGER GENERATED ALWAYS AS IDENTITY ( 
	START WITH 1 INCREMENT BY 1 
	NO MINVALUE NO MAXVALUE 
	NO CYCLE NO ORDER 
	CACHE 20 ) 
	, 
	CUSTOMER_NAME FOR COLUMN CUSTNAME   VARCHAR(30) CCSID 37 NOT NULL , 
	INSERT_TIMESTAMP FOR COLUMN WHENINS    TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP , 
	UPDATE_TIMESTAMP FOR COLUMN WHENUPD    TIMESTAMP GENERATED ALWAYS FOR EACH ROW ON UPDATE AS ROW CHANGE TIMESTAMP NOT NULL )   
	  
	RCDFMT CUSTOMERS  ; 


sql: SELECT * FROM THEBANK.CUSTOMERS;
-- Generated 1 insert statement

insert into THEBANK.CUSTOMERS (
  CUSTOMER_ID, CUSTOMER_NAME, INSERT_TIMESTAMP, UPDATE_TIMESTAMP
) values 
  (1, 'Scott', '2021-01-20 21:46:39.889538', '2021-01-20 21:46:39.889538'),
  (2, 'Sue', '2021-01-20 21:46:39.889995', '2021-01-20 21:46:39.889995');

-- here's what I would expect
insert into THEBANK.CUSTOMERS (
  CUSTOMER_NAME 
) values 
  ( 'Scott' ),
  ( 'Sue' );
```

janfh:
Option to exclude generated columns from sql-qualifier insert statements

Fixes #392 
Fixes #355

- Bumped mapepire-server to 2.3.3
- Bumped mapepire-js to 0.6.0
- New option `sqlIncludeGeneratedColumns`
- Remove autogenerated columns from sql insert statements if enabled

### DIFF
--- a/package.json
+++ b/package.json
@@ -152,6 +152,11 @@
               "Use the column name",
               "Use the column label\n'Extended metadata' must be set to true in the JDBC configuration"
             ]
+          },
+          "vscode-db2i.codegen.sqlIncludeGeneratedColumns": {
+            "type": "boolean",
+            "description": "Include generated columns in SQL insert statements",
+            "default": true
           }
         }
       },
--- a/src/views/results/contributes.json
+++ b/src/views/results/contributes.json
@@ -43,6 +43,11 @@
               "Use the column name",
               "Use the column label\n'Extended metadata' must be set to true in the JDBC configuration"
             ]
+          },
+          "vscode-db2i.codegen.sqlIncludeGeneratedColumns": {
+            "type": "boolean",
+            "description": "Include generated columns in SQL insert statements",
+            "default": true
           }
         }
       }
--- a/src/views/results/index.ts
+++ b/src/views/results/index.ts
@@ -403,7 +403,8 @@ async function runHandler(options?: StatementInfo) {
 
           setCancelButtonVisibility(true);
           updateStatusBar({executing: true});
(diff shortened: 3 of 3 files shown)


worksofliam:
@janfh This is marked as draft. Is it ready or shall I wait?

janfh:
> @janfh This is marked as draft. Is it ready or shall I wait?

I need to test some more. Not sure if we get enough info to exclude "generated always" (not auto increment). I actually think all columns should be included when using "select *". The data might be inserted into another table with another definition. What do you think?

worksofliam:
> I actually think all columns should be included when using "select *". The data might be inserted into another table with another definition.

I agree with this.

PR #409, opened 2025-06-09
Make columns resizable and improve Null values handling

### LINKED ISSUE #402: UX: No obvious way to enter null value when in update mode
When in update mode, for example:

`UPDATE:  SELECT * FROM SAMPLE.EMPLOYEE`

When we need to edit a null capable database column by changing its current value to a _null_, there is no obvious way to do so. Typing "null" with the keyboard will result in an alphanumeric string "null" being entered rather than _null_ flag/special value.

Perhaps a way to enter a _null_ already exists, but I was not successful finding and searching any documentation about DB2i extension beyond its README.md.

### LINKED ISSUE #403: UX: update mode output formatting does not refresh when _null_ value is changed
Having _null_ values displayed in _italic_ is great! (Perhaps a yellow or user-editable background color would be also a welcome feature.)

Example:
Image

.
.
.

But, when in edit mode, if a database field currently with _null_ value gets edited to something else, the _italic_ formatting related to the former _null_ value does not get refreshed to a regular style.

Example:
Image

dferrand:
Multiple changes to null value handling:
* In addition to being in italic, a null value now has a highlighted background (as suggested by @mipjones)
* When a field is updated from a null to a non null value, the italic and highlighted background is now removed (fixes #403)
* Setting a field to null is now done by pressing shift+enter instead of entering null (or - for a field shorter than 4 characters). A tooltip indicates this information if the field is nullable. This allows setting a field to the "null"string (or "-" if shorter than 4 characters). (fixes #402)

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,7 +5,7 @@
 [GitHub star this repo 🌟](https://github.com/codefori/vscode-db2i)
 [View our documentation 📘](https://codefori.github.io/docs/extensions/db2i/)
 
-Db2 for IBM i tools provides SQL functionality to VS Code. **Currently in preview**.
+Db2 for IBM i provides SQL functionality to VS Code.
 
 - Statement executor and result set view
 - Schemas view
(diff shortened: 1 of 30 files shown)


dferrand:
I made a few more changes:
* Fixed the css issues
* fixed the font
* column widths on result set are now adjustable. If you double click on the width adjustement handle, it will auto adjust the width to the content
* You can now choose to display both column name and label in the header

sebjulliand:
Note: I just met @dferrand who asked for the status of the PR. This is a bump to not forget it 😅

sebjulliand:
@dferrand can you resolve that small remaining conflict? Then I'll give it a go. Thanks!

dferrand:
Happy new year @sebjulliand !!

Sorry for the delay. I've made the changes and fixed the conflict.

sebjulliand (approved):
Excellent change @dferrand ; approved and merged 👏🏻

Issue #361, opened 2025-04-02
Please add key shortcuts for the new Run actions


forstie:
Hi Scott, 
Please give us a short cut to use the run the run all statements. 
My suggestion is command/control + A + R 
(combines the select all and run together) 
 
Thank you,
Marina

PR #417, opened 2025-06-24
Added default run all keybind #361

mkou-i:
For [#361](https://github.com/codefori/vscode-db2i/issues/361): 
- Customer suggested command/control + A + R
- VSCODE doesn't accept modifer + multiple keys. Tested locally, and it showed as 'unknown'. Documentation / editor allows for [key chords](https://code.visualstudio.com/docs/configure/keybindings#_accepted-keys). Settled on CTRL + a

### DIFF
--- a/src/views/results/contributes.json
+++ b/src/views/results/contributes.json
@@ -212,8 +212,8 @@
       },
       {
         "command": "vscode-db2i.runEditorStatement.multiple.all",
-        "key": "ctrl+r a",
-        "mac": "cmd+r a",
+        "key": "ctrl+alt+a",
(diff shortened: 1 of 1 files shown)


mkou-i:
Liam, Control/Command + A + R would be invalid. There are limitations to how many 'keys' can be used in one sequence. 

Changed it to a key chord -> "ctrl+r a". This would stray a bit away from the other key bindings.

SanjulaGanepola:
@mkou-i Ok so turns out that won't work either as hitting `ctrl+a` will also end up waiting for the next key. How about `ctrl+alt+a` / `cmd+alt+a`

PR #420, opened 2025-07-01
Support for SQL parameters and binding

worksofliam:
Adds a new prefix, `bind`, which allows users to bind values into prior statements that contain parameter markers or host variables.

```
select * from employee where workdept = ? and salary >= :minSalary;

bind: 'A00', 20000;
bind: 'A00', 40000;
```

You'll also find that when executing `bind` statements, history will be updated to store the original statement and will also include storing the `bind` statement being executed into the history of that statement.

Additionally, fixes some bugs for statement execution.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "vscode-db2i",
-  "version": "1.12.0",
+  "version": "1.13.3",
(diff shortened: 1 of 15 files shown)


forstie (approved):
Everything is working great now.

PR #422, opened 2025-07-17
Feature/dove graph

jonnyz32:
Introduce a new CytoscapeGraph class to manage graph elements and display them in a webview. Enable node selection to show detailed information.

Tasks:

 - [x] Merge in missing commits from main
 - [x] Fix that 'Run and explain' doesn't open the view, but 'Explain without running does'
 - [x] Improve formatting of nodes in webview
 - [x] stretch: Determine if coloring each node can be based on the old logic
https://github.com/codefori/vscode-db2i/blob/main/src/views/results/explain/doveTreeDecorationProvider.ts#L38
 - [x] stretch: See about embedding icons onto nodes
Here are the attached icons for each different node type: https://github.com/codefori/vscode-db2i/blob/main/src/views/results/explain/doveResultsView.ts#L13
Perhaps you need this to embed the icons into the webview? https://github.com/microsoft/vscode-codicons
 - [x] Attach description to each node if applicable
        Old logic here: https://github.com/codefori/vscode-db2i/blob/main/src/views/results/explain/doveResultsView.ts#L126
- [x] Stop using remote URL to pull in cytoscape and embed into the build
- [x] Update icon.ts to include all possible labels

https://github.com/user-attachments/assets/1b38fca1-3563-4dcc-990d-0ce63f0ad154


worksofliam (changes requested):
The UX is looking really good, but there are some things in the code we'll need to cleanup before merge.

### context

We don't typically want to store the extension context as a global, and we usually want to pass it all the way down where possible. The route I would take is to start by deleing the `contextProvider.ts` file and seeing how you can mangle it down there.

### build files (`media`)

We add a new dependency in this PR - cytoscape - but, we should not be checking in this file into our git repository. Instead, at build time, we should be pulling it from the installed `node_modules.`

The remaining files make sense, since it's written outside of a package and is part of the extension.

SanjulaGanepola:
Closing this PR as new visual explain graph PR is being worked on

PR #425, opened 2025-08-05
Fix comment for multiline statement in generated RPG

janfh:
When using rpg-qualifier on multiline statement, the statement was inserted as a comment with "//" only on the first line.


worksofliam (approved):
Tested with the following statement and it works good!

```sql
rpg: select * from 
sndbx87.employee;
```

Issue #169, opened 2023-11-17
Feature: Import


worksofliam:
I'd like the ability to import different formats: xml, JSON array of objects and CSV with optional title.

It wouldn't import directly into a table, it would actually generate the SQL insert statement from the data provided.

Optionally, be really cool and generate a basic `create` statement for data, where the data types are determined based on the data imported.

PR #435, opened 2025-08-27
Create INSERT statement based on imported csv or json

julesyan:
The command is "Generate Insert from File". Tested with the attached files, including with blanks. Will open a new file with the newly generated insert statement. 

Closes issue #169 

[export.csv](https://github.com/user-attachments/files/22007134/export.csv)
[export.json](https://github.com/user-attachments/files/22007135/export.json)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "vscode-db2i",
-  "version": "1.13.3",
+  "version": "1.14.0",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "vscode-db2i",
-      "version": "1.13.3",
+      "version": "1.14.0",
       "dependencies": {
         "@ibm/mapepire-js": "^0.5.0",
         "@octokit/rest": "^21.1.1",
@@ -15,6 +15,7 @@
         "json-to-markdown-table": "^1.0.0",
         "lru-cache": "^6.0.0",
         "node-fetch": "^3.3.1",
+        "octokit": "^5.0.3",
         "showdown": "^2.1.0",
         "sql-formatter": "^14.0.0"
       },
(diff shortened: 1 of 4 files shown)


forstie:
Looking at the code is aok, but trying the function is better.
Can I please get a vsix?

forstie:
If the spreadsheet does not include the column names as the first row, the generated INSERT won't work.
Is there a way to handle this situation or does this need to be pointed out in the documentation?
image

forstie:
If we could detect that the 1st row wasn't simple names, we could generate an INSERT with VALUES statement that simply avoided the column names syntax part. That piece is optional but would infer that the spreadsheet was generated by a SELECT * FROM xyz type of query.

julesyan:
[vscode-db2i-1.14.0.vsix 3.zip](https://github.com/user-attachments/files/22011076/vscode-db2i-1.14.0.vsix.3.zip)

@forstie based on our discussion i now prompt the user if the csv has headers or not, it will generate based on that answer

julesyan:
[vscode-db2i-1.14.0.vsix.zip](https://github.com/user-attachments/files/22011361/vscode-db2i-1.14.0.vsix.zip)
 Changed it so any file will show the right click menu. If the file is not csv or json it will prompt the user

julesyan:
[vscode-db2i-1.14.0.vsix.zip](https://github.com/user-attachments/files/22011854/vscode-db2i-1.14.0.vsix.zip)

Allowed column names to contain spaces

julesyan:
[vscode-db2i-1.14.0.vsix.zip](https://github.com/user-attachments/files/22044117/vscode-db2i-1.14.0.vsix.zip)

SanjulaGanepola (approved):
Looks good! Nice work

PR #448, opened 2025-09-16
Updated keyboard shortcuts for explaining and running

sebjulliand:
This PR changes the default keyboard shortcuts for `Explain` and `Run and Explain` action so they don't overlap with other shortcuts like `Run in new view`.

image

- `Explain` is `ctrl+e`
- `Run and Explain`  is `ctrl+u`
- `Run from cursor`  is `ctrl+shift+r`
- `Run all`  is `ctrl+shift+a`

This is consistend with ACS's keybindings.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1459,6 +1459,18 @@
         "when": "editorLangId == sql"
       },
       {
+        "command": "vscode-db2i.runEditorStatement.multiple.all",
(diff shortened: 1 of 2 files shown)


sebjulliand:
Thanks @chrjorgensen!

I added two more key bindings:
image

What do you think?

PR #459, opened 2025-10-10
Add `udtf` SQL prefix

SanjulaGanepola:
### Changes

This PR adds `udtf` as a new SQL prefix for generating the skeleton for a user-defined table function as requested in https://github.com/codefori/vscode-db2i/issues/458

image

### DIFF
--- a/src/views/results/codegen.ts
+++ b/src/views/results/codegen.ts
@@ -79,8 +79,8 @@ export function queryResultToUdtf(result: QueryResult<any>, sqlStatement: string
     tokenIs(tokens[1], `colon`, `:`) &&
     tokenIs(tokens[2], `statementType`, `SELECT`) &&
     tokenIs(tokens[3], `asterisk`, `*`)) {
-    const prefixEnd = tokens[3].range.start - tokens[0].range.start - tokens[1].range.start - 2;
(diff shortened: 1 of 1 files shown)


forstie:
I think I have the vsix properly updated. 
I am seeing...

udtf: select * from dogshow.account;

Create this, which is wrong

-- User-defined table function
CREATE OR REPLACE FUNCTION MyFunction()
  RETURNS TABLE (
    CLINAME VARCHAR(100)
  )
  NOT DETERMINISTIC
  NO EXTERNAL ACTION
  READS SQL DATA
  SET OPTION COMMIT = *NONE,
             DYNUSRPRF = *USER,
             USRPRF = *USER
  BEGIN
    RETURN CLINAME
            select * from dogshow.account;
  END;

forstie:
I did the install over again, cmd-Q to kill vs, restarted and still see this
image

SanjulaGanepola:
@forstie I gave this a try again and it works fine on my end. I also had Devansh try this out and it also works for him (he has a Mac). Can you give this a try once more?

@chrjorgensen Would you be able to give this a try?

forstie (changes requested):
I added a comment with a query that doesn't generate cleanly using the new udtf: keyword.
Please take a look at it.

PR #467, opened 2025-11-13
fix for #466 Update status could also be UP_TO_DATE

### LINKED ISSUE #466: Missing UpdateStatus.UP_TO_DATE Check Causes "Database server component is required" Error
**Describe the bug**

When using the exported API of the db2i extension to create a new job immediately after the connected event from the code-for-ibmi extension, the following error occurs:

```
Database server component is required. Please see documentation for details.
```

If I introduce a delay before creating the job, the error does not occur. Upon investigation, it seems the issue lies in the db2i extension's handling of the server component update status. The current implementation only checks for `UpdateStatus.JUST_UPDATED` but does not account for `UpdateStatus.UP_TO_DATE`.

**To Reproduce**

1. Use the code-for-ibmi extension to connect to an IBM i server.
2. Immediately after the connected event, use the db2i API to create a new job.
3. Observe the error: Database server component is required. Please see documentation for details.

**Actual Behavior**
The db2i extension only checks for UpdateStatus.JUST_UPDATED, leading to the error when the server component is already up to date but not explicitly marked as just updated.

**Expected behavior**

The db2i extension should handle both UpdateStatus.JUST_UPDATED and UpdateStatus.UP_TO_DATE to ensure the server component is ready before proceeding.

drahreg:
PR which should fix #466

### DIFF
--- a/src/connection/manager.ts
+++ b/src/connection/manager.ts
@@ -166,7 +166,7 @@ export class SQLJobManager {
 
     } else if (!ServerComponent.isInstalled()) {
       let updateResult = await ServerComponent.checkForUpdate();
-      if (UpdateStatus.JUST_UPDATED === updateResult) {
+      if (UpdateStatus.JUST_UPDATED === updateResult || UpdateStatus.UP_TO_DATE === updateResult) {
         await onConnectOrServerInstall();
         return this.getPagingStatement(query, opts);
       }


sebjulliand (approved):
Fixed and approved live.
Thanks @drahreg.

PR #476, opened 2025-12-10
added return in vscode-db2i.dove.export

ibm-Abhinav:
added return statement inside vscode-db2i.dove.export to get visual explain json and use it in mcp tool for visual explain.


venky225:
We can close this PR as we are following a different approach for Visual explain

PR #480, opened 2025-12-23
Add exported API to retrieve SQL examples

SanjulaGanepola:


worksofliam:
@SanjulaGanepola I wonder if this needs to be part of the extension API vs a command?

forstie (approved):
Works great!

PR #482, opened 2026-01-14
Fix actions on quoted database objects

SanjulaGanepola:
This PR fixes actions performed on quoted schemas / database objects:
- `Clear`
- `Copy Data`
- `Delete`
- `Rename`
- `View Permissions`
- `Advised Indexes`
- `Get Related Objects`
- `Get Indexes`
- Get MTIs
- `Get Authorities`
- `Get Object Locks`
- `Get Record Locks`

### DIFF
--- a/src/database/schemas.ts
+++ b/src/database/schemas.ts
@@ -3,6 +3,7 @@ import path from "path";
 import { getInstance } from "../base";
 import { JobManager } from "../config";
 import { ResolvedSqlObject, BasicSQLObject } from "../types";
+import Statement from "./statement";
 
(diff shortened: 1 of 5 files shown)


forstie:
@SanjulaGanepola the vsix confirmed that things are much better.
I did notice two actions that still don't handle delimited names.
On COMMON1, under schema "annoying", the following actions still do not work.

1) Table "annoyingly123", the CLEAR action should use "annoy1" not annoy1.
2) Copy Data - similar kind of problem, need to fill in the system names into the command, but must user QSYS2.DELIMIT_NAME() to make sure the name is correct

forstie (approved):
The myriad of changes will make a huge difference to anyone with delimited names, thank you.
In my testing, I found 2 things that still didn't work.
I'll leave it to you to decide if you want to fix those.

PR #483, opened 2026-01-14
Add `Collapse All` button to all tree views

SanjulaGanepola:
The built-in `Collapse All` button is now available on all tree views:
image


sebjulliand (approved):
Looks good 👍🏻

PR #484, opened 2026-01-14
Save update prefix statement to statement history

### LINKED ISSUE #462: Update prefix suggestion
With the Update: prefix, it would be very beneficial to be able to have the update statements the were ran behind the scenes accessible after the fact.  This would allow an initial check of "what did I just run?".   But this could allow the user to be able to save off what they ran incase the have to replicate it over again for testing or something.

SanjulaGanepola:
This PR adds the executed update statement to the `Statement History` view. Note that the `?` markers are kept as is. This is consistent with how the current `bind` prefix adds to the history.

Fixes #462

image

### DIFF
--- a/src/views/results/resultSetPanelProvider.ts
+++ b/src/views/results/resultSetPanelProvider.ts
@@ -13,7 +13,7 @@ import { TableColumn } from "../../types";
 import { statementDone } from "./editorUi";
 import { QueryResult } from "@ibm/mapepire-js";
 
-export type SqlParameter = string|number;
+export type SqlParameter = string | number;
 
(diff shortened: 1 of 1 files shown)


sebjulliand (approved):
Works as advertised; nice addition 👍🏻 
Merge whenever you want.

forstie (changes requested):
While its absolutely correct that the UPDATE now appears in the statement history, the users requesting this support are going to expect to see a little bit more.
Can you consider adding the parameter marker values as comments?

UPDATE toystore.sales t SET t.SALES = ? WHERE SALES_ID = ?;

PR #492, opened 2026-01-22
Add support for setting any default JDBC option and setting default saved configurations

### LINKED ISSUE #491: Default Connection Settings
When defining an IBMi system to connect to, I would like an option to specify the database (iASP) name I want to use in my initial connection. The only 2 ways I can see to get into the iasp after making the connection is to modify the current connection settings from the default job or go into the DB2i view , create a new connection under SQL Job Manager which has the option to specify a Database Name and use that config instead. This would have to get started manually though as this config is not referenced upon the initial connection to the IBMi server. I think there are other opportunities to include some additional default settings as well like SQL vs System Naming and setting your Library list etc. Thanks - Rich

SanjulaGanepola:
### Changes
1. Adds support for setting any default JDBC option (except `libraries`) instead of just SQL Naming. There is also logic to migrate what users may have previously set in the old `Job Naming Default` setting. These default connection properties are not specific to any connection and will be used to spin up a job (if a connection has no default saved configuration - this is also a new feature explained further below). This addresses a customer reported issue (during the Code for IBM i Fridays) where you could not change any JDBC options unless you connected at least once (this is now possible).
    image
2. Adds support for setting default saved configurations so whenever you connect, it will spin up a job with those configurations. This addresses https://github.com/codefori/vscode-db2i/issues/491
    image
    image
3. When spinning up a job, it often felt like it wasn't doing anything unless you were looking at the status bar. I also added a progress bar to the view itself so users knows it's spinning up.
    image


forstie (approved):
Works like a champ!

PR #502, opened 2026-02-18
Fixed schema browser issues

sebjulliand:
This PR fixes two minor issues in the Schema Browser:
- The `Get locks` and `Get authorities` actions could not return any result for objects other than tables. Using the object/schema names instead of the system's fixes the issue.
- Functions with anonymous parameters could not be expanded (it would throw an error. For example, this kind of function:
  ```sql
  CREATE or replace FUNCTION SJULLIAND.ANONYMOUS_PARMS ( 
	INTEGER, INTEGER, INTEGER ) 
	RETURNS SMALLINT CAST FROM CHAR(1)   
	LANGUAGE RPGLE 
	SPECIFIC SJULLIAN.ANYMSPARMS
	NOT DETERMINISTIC 
	NO SQL 
	RETURNS NULL ON NULL INPUT 
	NO EXTERNAL ACTION 
	ALLOW PARALLEL 
	EXTERNAL NAME MYSRVPGM(MYPROC) 
	PARAMETER STYLE GENERAL ;
  ```
  
  image

- The code was cleaned up a bit for clarity (use Maps instead of Records, marked fields as readonly
- Delete, Rename, Get Locks and Get Authorities were added to the `Logical` items.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1355,12 +1355,12 @@
         },
         {
           "command": "vscode-db2i.getAuthorities",
-          "when": "viewItem == table || viewItem == view || viewItem == alias || viewItem == function || viewItem == variable || viewItem == index || viewItem == procedure || viewItem == receiver || viewItem == journal || viewItem == sequence || viewItem == package || viewItem == trigger || viewItem == type",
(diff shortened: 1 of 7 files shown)


forstie:
While testing your changes, I noticed that there are no actions for logical files.
It seems like we have actions that should work fine on those types of objects.
COMMON76 - TOYSTORE3/ITEMLF1  is an example

image

forstie (approved):
Tested well.
I noticed something while testing and left a comment.

PR #506, opened 2026-03-03
Added Clear Results button and action

### LINKED ISSUE #505: Ability to clear output panel results
Is there going to be an easy way to clear the IBM i query Results output panel results coming?

Today if I stop showing the Results view in the IBM i output panel the results of my last query are cleared but this does not seem intuitive. 

There are times when I have to query tables with sensitive values.  I know in those cases I could just use ACS RSS product but I would like to not go there is I could. 

Thanks, 
Matt

sebjulliand:
# Description
Resolves #505 

This PR adds a `Clear Results` button in the Result Set view title toolbar.
image

It is also available from the command palette.
image

The action/button is only enabled when there are results to clear.

Running this action does what it says on the tin: it clears the view entirely.

# How to test
1. Run a query
2. The Clear Results button must become enabled
3. Click on it: the results should be cleared and the button disabled

### DIFF
--- a/package.json
+++ b/package.json
@@ -685,6 +685,13 @@
         "icon": "$(gear)"
       },
       {
+        "command": "vscode-db2i.resultset.clear",
+        "title": "Clear Results",
+        "category": "Db2 for i",
+        "icon": "$(clear-all)",
+        "enablement": "vscode-db2i:canClear"
+      },
+      {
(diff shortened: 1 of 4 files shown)


forstie (approved):
Works like a champ. Approved!
Way2go

PR #512, opened 2026-03-19
Add rpgle-repl to Notebook

tom-writes-code:
**Expose [rpgle-repl ](https://github.com/tom-writes-code/rpgle-repl#)to Notebooks.**

DB2 i notebooks already allow inline SQL code snippets to be executed on the fly.

This functionality allows for small RPG examples to be included too, which could be used to confirm behaviour of %BiFs, or for simple training examples.

RPGLE-REPL was changed to allow code to be passed in to an SQL function, which then runs the REPL process, and returns a data set which can be parsed back to the user.

Setting for where RPGLEREPL lives (blank by default)

image

Option to change language mode

image

No connection, same as SQL

image

Configuration not installed

image

Compilation failure

image

Successful run

image

### DIFF
--- a/RPGLE-NOTEBOOK-PLAN.md
+++ b/RPGLE-NOTEBOOK-PLAN.md
@@ -1,0 +1,681 @@
+# RPG Notebook Integration — Implementation Plan
+
+## Executive Summary
+
+Add RPG cell support to IBM i Notebooks (vscode-db2i) using RPGLE-REPL as the
+backend execution engine. Users write RPG snippets in notebook cells, the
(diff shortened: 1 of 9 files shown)


tom-writes-code:
The thing I'd love to change but can't work out where it comes from is the syntax highlighting - I'd love it to treat column 1 as the start of the line, rather than column 7 or 8.

image

SanjulaGanepola:
@tom-writes-code This looks like a cool enhancement, but is not a fit for this extension. Also it requires an external dependency to be installed which we tend to avoid in most cases. However, I would definitely suggest putting this into your own extension.

PR #514, opened 2026-04-11
Fix conflicting `Explain without running` keyboard shortcut

### LINKED ISSUE #503: RunSQL statement not respecting Code4i user library list
I have a sql script that:
*  creates table, 
* inserts labels on columns
* assigns permissions to that table. 

I used to be able to connect to Code4i, open the script and hit Ctl+E to basically compile (run) all three statements.  Everything worked great.  My table was created and permissions set.

Something has changed.  When I hit Ctl+E it seems like it does try to run one statement, I did find the "Run all" but when I run the sql statements now it does not respect my user library list in code4i ??

How can I run sql code in vscode and not have to hard code libraries now??

SanjulaGanepola:
### Changes

This PR changes the `Explain without running` keyboard shortcut to `ctrl+alt+u` / `cmd+alt+u` because the current `ctrl+e` / `cmd+e` conflicts with Code4i's `Run Action` command which is more commonly used.

Fixes #503

image

### How to Test
1. Use `ctrl+e` to verify Code4i's `Run Action` works
2. Use `ctrl+alt+u` to verify `Explain without running` works

### DIFF
--- a/package.json
+++ b/package.json
@@ -1635,8 +1635,8 @@
       },
       {
         "command": "vscode-db2i.editorExplain.withoutRun",
-        "key": "ctrl+e",
-        "mac": "cmd+e",
+        "key": "ctrl+alt+u",
+        "mac": "cmd+alt+u",
         "when": "editorLangId == sql && resourceExtname != .inb"
       },
       {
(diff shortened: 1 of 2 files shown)


forstie (approved):
I see your point. Approved

PR #517, opened 2026-04-19
Fix `UDTF` prefix bugs

### LINKED ISSUE #498: `udtf` prefix does not handle UDTs
Found by @forstie:

Image

SanjulaGanepola:
### Changes

This PR address the following 2 issues:
- There was a bug with generating a UDTF if there were multiple spaces after the prefix (ie. like below with 2 spaces after):
  ```sql
  udtf:  select * from sample.employee
  ```
- Fixes https://github.com/codefori/vscode-db2i/issues/498 - Now supports UDTs
  ```sql
  udtf: select * from fourty.many_data_types;
  ```


forstie (approved):
Works like a champ. Nice going

PR #518, opened 2026-04-19
Add support for `md` prefix to generate markdown tables

SanjulaGanepola:
### Changes

This PR adds support for a `md` prefix to generate tables in markdown format.

image


forstie:
Hi,
So, after a simple test worked, I tried this:

md: Select * from qsys2.sysixadv order by times_advised desc limit 1;

Is this what you are expecting to have returned?
image

SanjulaGanepola:
@forstie Yes that should be correct. The reason it doesn't look readable on your end is because you have word wrap enabled.

You can toggle this in the command pallete:

image

Should look something like this where you can scroll to the right:

image

You can also then use VS Code's buil-tin preview option:

image

to preview a rendering of the markdown table:

image

forstie (approved):
perfecto

PR #519, opened 2026-04-19
Exclude column label if same as column name

SanjulaGanepola:
### Changes

This PR excludes the column label if it is the same as the column name

image


forstie (approved):
works like a champ!
Thanks

PR #522, opened 2026-04-22
Add statements with prefixes to the history

SanjulaGanepola:
### Changes
1. Added `json`, `csv`, `cl`, `md`, `sql`, `rpg`, `udtf` prefixes to the set of prefixes that will be saved in the `Statement History` view
2. Added `rpg` and `udtf` prefixes (so they are consistent with the others) to the set of allowed prefixes that can be used when running multiple statements

### DIFF
--- a/src/views/results/index.ts
+++ b/src/views/results/index.ts
@@ -59,6 +59,12 @@ let doveResultsTreeView: TreeView<ExplainTreeItem> = doveResultsView.getTreeView
 let doveNodeView = new DoveNodeView();
 let doveNodeTreeView: TreeView<PropertyNode> = doveNodeView.getTreeView();
 let doveTreeDecorationProvider = new DoveTreeDecorationProvider(); // Self-registers as a tree decoration provider
+const ALLOWED_PREFIXES_FOR_MULTIPLE: StatementQualifier[] =
+  [`json`, `csv`, `md`, `cl`, `sql`, `rpg`, `udtf`, `statement`, `bind`];
+const ALLOWED_PREFIXES_FOR_HISTORY: StatementQualifier[] =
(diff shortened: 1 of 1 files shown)


src/views/results/index.ts
```diff
@@ -59,6 +59,12 @@ let doveResultsTreeView: TreeView<ExplainTreeItem> = doveResultsView.getTreeView
 let doveNodeView = new DoveNodeView();
 let doveNodeTreeView: TreeView<PropertyNode> = doveNodeView.getTreeView();
 let doveTreeDecorationProvider = new DoveTreeDecorationProvider(); // Self-registers as a tree decoration provider
+const ALLOWED_PREFIXES_FOR_MULTIPLE: StatementQualifier[] =
+  [`json`, `csv`, `md`, `cl`, `sql`, `rpg`, `udtf`, `statement`, `bind`];
+const ALLOWED_PREFIXES_FOR_HISTORY: StatementQualifier[] =
```
forstie, line 64: should update be included in this list?

forstie (commented):
question for you

src/views/results/index.ts
SanjulaGanepola: `update` has different logic where we don't actually save the statement with the prefix. Rather we store whatever updates are made. Example:

```
update: select * from toystore.employee;
```

ends up strong the following as we make updates:

```
UPDATE toystore.employee t SET t.FIRSTNME = ? WHERE RRN(t) = ?;

-- Substatements: 4
bind: 'CHRISTINE', 1;
bind: 'MICHAE', 2;
bind: 'MICHAEL', 2;
bind: 'CHRISTINEA', 1;
```

forstie (approved):
bueno

PR #523, opened 2026-04-23
Fix syntax errors impacting parsing of subsequent statements

SanjulaGanepola:
### Changes
Although syntax errors can definitely impact parsing, there are cases where we can gracefully handle parsing of subsequent lines especially when a user has a semi-colon.

#### Previously
If there was a line with a syntax error (such as extra brackets in the example below), it flagged the syntax errors. However, subsequent lines were not being parsed correctly.

image

We also wouldn't be able to run the next statement correctly to create the procedure as the statement range is incorrectly parsed:
image

#### Now
The bug in parsing was that the `bracketDepth` was not being reset as there is a bracket mismatch. However, since the user has a semi-colon, we can gracefully treat that as the end of the statement and thus reset the count to move onto parsing the next statement as normal.

We now only get syntax errors for the first line as expected and can run the create procedure statement.

### DIFF
--- a/src/language/sql/document.ts
+++ b/src/language/sql/document.ts
@@ -59,6 +59,7 @@ export default class Document {
           this.addStatement(statementTokens);
 
           statementStart = i + 1;
+          bracketDepth = 0;
           break;
 
         case `statementType`:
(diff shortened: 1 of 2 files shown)


forstie (approved):
Works well

PR #529, opened 2026-04-29
feat: Add "Copy as SQL List" context menu commands for SQL editors

bobcozzi:
## Summary

Adds two new right-click context menu commands to SQL editors, available whenever text is selected. This addresses a feature request from the IBM i OSS community where users wanted a quick way to convert a plain newline-separated list of values (e.g. product codes, order numbers) into a SQL-friendly quoted, comma-separated list for use in `IN` clauses and similar constructs — without leaving VS Code to use another tool like dBeaver.

> Community request: https://chat.ibmioss.org/#narrow/channel/8-vscode/topic/Seb's.20Extension.20Example.20Part.20deux/with/4188

---

### `Db2 for i: Copy as SQL List`
Instantly copies the selected newline-separated values to the clipboard using the user's current settings. No prompt — fast path for users who code quickly and want a single keystroke-style experience.

### `Db2 for i: Copy as SQL List (Prompt Format)`
Shows a QuickPick menu with format options before copying. When the selection contains numeric values, additional unquoted variants are offered:

| Option | Example output |
|---|---|
| Copy as list | `'S10_1678', 'S10_1949', 'S10_2016'` |
| Copy as list (numbers unquoted) | `'ABC', 123, 'DEF'` |
| Copy wrapped in parentheses | `('S10_1678', 'S10_1949', 'S10_2016')` |
| Copy wrapped in parentheses (numbers unquoted) | `('ABC', 123, 'DEF')` |

Both commands only appear in the context menu and command palette when a SQL file has an active text selection (`editorLangId == sql && editorHasSelection`).

---

## New Settings

| Setting | Type | Default | Description |
|---|---|---|---|
| `vscode-db2i.delimitedList.quoteNumbers` | boolean | `true` | Whether the instant-copy command quotes numeric values |
| `vscode-db2i.delimitedList.wrapInParentheses` | boolean | `false` | Whether the instant-copy command wraps the list in parentheses |

---

## Implementation Notes

- **SQL-safe quoting**: Internal single quotes are escaped as `''` (e.g. `O'Brien` → `'O''Brien'`)
- **Blank line filtering**: Empty/whitespace-only lines in the selection are silently ignored
- **Large selection guard**: A modal confirmation is shown if more than 256 values are selected
- **Clipboard error handling**: `clipboard.writeText()` failures surface a descriptive error message rather than failing silently
- New file: `src/language/clipboard.ts` — follows the same `initialise(context)` pattern as `src/language/json.ts`

### DIFF
--- a/package.json
+++ b/package.json
@@ -79,6 +79,16 @@
               "Java 17",
               "Java 21"
             ]
+          },
+          "vscode-db2i.delimitedList.quoteNumbers": {
+            "type": "boolean",
+            "description": "Quote numeric values when copying a selection as a SQL list",
+            "default": true
+          },
+          "vscode-db2i.delimitedList.wrapInParentheses": {
+            "type": "boolean",
+            "description": "Wrap the SQL list in parentheses when using 'Copy as SQL List' (instant copy)",
+            "default": false
           }
         }
       },
@@ -540,6 +550,16 @@
       {
         "command": "vscode-db2i.importDataContextMenu",
         "title": "Generate Insert from File",
+        "category": "Db2 for i"
+      },
+      {
+        "command": "vscode-db2i.copyAsDelimitedList",
+        "title": "Copy as SQL List",
+        "category": "Db2 for i"
+      },
+      {
+        "command": "vscode-db2i.copyAsDelimitedListPrompt",
+        "title": "Copy as SQL List (Prompt Format)",
         "category": "Db2 for i"
       },
       {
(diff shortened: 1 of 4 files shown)


buzzia2001 (approved):
works as expected, thank you @bobcozzi

sebjulliand:
Nice feature @bobcozzi ; don't forget to add it to the documentation so it doesn't become a hidden gem 😉 
Thanks!

bobcozzi:
> Nice feature @bobcozzi ; don't forget to add it to the documentation so it doesn't become a hidden gem 😉 Thanks!

Done! https://github.com/codefori/docs/pull/109

PR #539, opened 2026-05-25
Component update

### LINKED ISSUE #524: improve security
- Running CL commands should be qualified by the library (runsqlstm, clrpfm, cpyf)
- Objects created on the IBM i should be owned by a system id 
- Version checking of objects on the IBM i should incorporate a hash function
- UDTF should not use adopted authority. e.g. use EXEC SQL SET OPTION USRPRF=*USER, DYNUSRPRF=*USER

### LINKED ISSUE #535: mapepire server - use code4i to install it
Code4i installs the mapepire server so if Db2i uses Code4i then it doesn't need to install the mapepire server itself.

### LINKED ISSUE #544: Syntax checker is not working.
After the last update the syntax checker on SQL querys is not working either on SQL Documents or Notebooks (inb), every time i write a new word the next error is displayed 

Image

The execution of the query still working perfectly it is just the syntax checker  that is failing

### LINKED ISSUE #546: Edit connection properties button for SQL job manager does not open connection properties
**Describe the bug**

**To Reproduce**

- Connect to an IBM i system
- Click on the "edit connection properties" button

**Expected behavior**

The "edit connection properties" button should open the connection properties window

**Screenshots**

The following errors appear in the developer console:

Image

**Environment**

 - OS: Mac OS 15.7.3
 - Extension Version: 1.17.0

Debug information is below:

`<html><body bgcolor="white">

<hr>
<b>[DATASTREAM_IN]: </b><i>2026-06-09.16.49.49.414</i>
<font color="DarkSlateGrey">
<blockquote>
<pre>
{"id":"id2","type":"connect","technique":"tcp","application":"vscode-db2i 1.17.0","props":"naming=system;full open=false;transaction isolation=none;query optimize goal=1;block size=512;date format=iso;extended metadata=true;libraries=QGPL,QTEMP"}
</pre>
</blockquote>
</font>

<hr>
<b>[DATASTREAM_IN]: </b><i>2026-06-09.17.00.37.265</i>
<font color="DarkSlateGrey">
<blockquote>
<pre>
{"id":"id3","type":"setconfig","tracedest":"IN_MEM","tracelevel":"DATASTREAM"}
</pre>
</blockquote>
</font>

<hr>
<b>[DATASTREAM_OUT]: </b><i>2026-06-09.17.00.37.269</i>
<font color="DarkSlateBlue">
<blockquote>
<pre>
{"id":"id3","tracedest":"IN_MEM","tracelevel":"DATASTREAM","jtopentracedest":"IN_MEM","jtopentracelevel":"OFF","success":true,"execution_time":0}
</pre>
</blockquote>
</font>

<hr>
<b>[DATASTREAM_IN]: </b><i>2026-06-09.17.01.00.823</i>
<font color="DarkSlateGrey">
<blockquote>
<pre>
{"id":"id4","type":"gettracedata"}
</pre>
</blockquote>
</font>

<hr>
<b>[INFO]: </b><i>2026-06-09.17.01.00.823</i>
<font color="black">
<blockquote>
<pre>
synchronously dispatcing a request of type: GetTraceData
</pre>
</blockquote>
</font>
</body></html>`

julesyan:
Update to match the component checking in base extension v3.0.0


julesyan:
Putting in draft until v3.0.0 is out

sebjulliand:
We should make a note to remove the Mapepire Component from this extension and use the core's now.
We'd need to add support for the JDBC options in the core, but otherwise, it should be fine.

src/connection/components/checkStatement.ts
```diff
@@ -0,0 +1,91 @@
...
+    return {
+      status: remoteSignature ? "Installed" : "NotInstalled",
+      remoteSignature: remoteSignature,
+    };
+  }
+
+  /**
+   * Custom implementation that uses EXTERNAL_NAME instead of ROUTINEDEF
+   */
+  private async getSQLRoutineSignature(connection: IBMi, library: string, name: string, type: "PROCEDURE" | "FUNCTION") {
```
SanjulaGanepola, line 42: Isn't this duplicate of what is in `IBMiContent#getSQLRoutineSignature` so we should use that API?

src/connection/components/checkStatement.ts
```diff
@@ -0,0 +1,91 @@
...
+
+  async getRemoteState(connection: IBMi, installDirectory: string): Promise<SecureComponentState> {
+    const remoteSignature = await this.getSQLRoutineSignature(
+      connection,
+      connection.getConfig().tempLibrary.toUpperCase(),
+      CheckStatementComponent.FUNCTION_NAME,
+      CheckStatementComponent.TYPE,
+    );
+    return {
+      status: remoteSignature ? "Installed" : "NotInstalled",
```
SanjulaGanepola, line 34: Shouldn't we make use of the old `getVersionOf` still so we can still make use of the `NeedsUpdate` state?

SanjulaGanepola (changes requested):

src/connection/components/checkStatement.ts
julesyan: this one uses "EXTERNAL_NAME" instead of "ROUNTINEDEF" since this function doesnt have a routine definition

src/connection/components/checkStatement.ts
julesyan: There is no guarantee the long comment is correct, this verison we grab is either not installed or is installed

julesyan:
image

I get this when trying to start any SQL job (saved configuration or through the New SQL Job button). I cant run any commands without starting an SQL job. 

Is it expected that the SQL Job Manager still exists in Db2i if the job we use is supposed to be the one from Core? If not then does that mean i lose all my configurations?

sebjulliand:
No, you're not supposed to lose the configuration.
And the Core uses its own job. Tehe Db2 extension will create and use one of its own (and several others, from the manager).
So yes, the manager is needed to monitor and manage the SQL jobs started from the Db2 extension.

I didn't get that error when testing it. That may come from existing configurations (besides, there was a lot of TypeScript type cheking that needed to be fixed - this may have slipped in).

To locate where the issue comes from, enable these two in VS Code debug view:
image

And then try to start an SQL job.

julesyan:
image

Looks to be from this line: https://github.com/codefori/vscode-db2i/blob/954ce675d0c028da04a6799bc949d253cdbb7775/src/connection/manager.ts#L53

julesyan:
ok i ran it again using 3.0.9 and it works 😄

sebjulliand:
> ok i ran it again using 3.0.9 and it works 😄

Yes, I just figured it out looking at your call stack 😅 Phew!

julesyan:
Not sure if this is new but when I spawn my "Read only" job (I checked the config it is correct), it does not have the same configuration, specifically it has access set to All. Tested and i can delete data with this job 

Edit: when i set it to read only it does not update the job, I can still delete data

sebjulliand:
Code for IBM i 3.0.10 needs to be released to fix this.

julesyan:
Tested and working from my end, @SanjulaGanepola can you review as I am the author of the PR so I cant approve it

src/views/jobManager/editJob/otherTab.ts
```diff
@@ -367,21 +367,9 @@ export default function getOtherTab(options: JDBCOptions) {
...
-        description: `False`,
-        selected: options["secure"] !== true,
-      },
-      {
-        value: `true`,
-        text: `encrypt all client/server communication`,
-        description: `True`,
-        selected: options["secure"] === true,
-      },
-    ])
```
SanjulaGanepola, line 384: Just for my own understanding, why was this option removed?

src/language/providers/logic/parse.ts
```diff
@@ -1,6 +1,6 @@
 import { TextDocument } from "vscode";
 import Document from "../../sql/document";
-import { VALID_STATEMENT_LENGTH } from "../../../connection/syntaxChecker/checker";
+import { VALID_STATEMENT_LENGTH } from "../../../connection/components/validateStatement";
 
 let cached: Map<string, {ast, version}> = new Map();
```
SanjulaGanepola: To avoid:

Image

```suggestion
let cached: Map<string, {ast: Document, version: number}> = new Map();
```

src/connection/extendedSQLJob.ts
```diff
@@ -0,0 +1,60 @@
...
+    const connection = getInstance().getConnection();
+
+    // Note that this statement is run via the base extension since it has to be done on a job other than the one whose SQL is getting canceled
+    await connection.runSQL(`CALL QSYS2.CANCEL_SQL('${wrappedJob.id}')`);
+
+    const [row] = await connection.runSQL(`select V_SQL_STMT_STATUS as STATUS from table(qsys2.get_job_info('${wrappedJob.id}'))`) as { STATUS: string | null }[];
+    return !(row && row.STATUS === `ACTIVE`);
+  }
+  return wrappedJob;
+}
```
SanjulaGanepola, line 60: Since it looks like we are actually extending the `SQLJob` class, why not actually create a class and extend it?

Something like:
```ts
import { SQLJob } from "@ibm/mapepire-js";
import { QueryResult } from "@ibm/mapepire-js/dist/src/types";
import { getInstance } from "../base";
import { SelfValue } from "../views/jobManager/selfCodes/nodes";
import { JobLogEntry } from "./types";
import { NamingFormats } from "./manager";

export class ExtendedSQLJob extends SQLJob {
  private selfState: SelfValue = "*NONE";
  private currentSchemaStore: string | undefined;

  getSelfCode(): SelfValue {
    return this.selfState;
  }

  resetCurrentSchemaCache() {
    this.currentSchemaStore = undefined;
  }

  async getCurrentSchema(): Promise<string> {
    if (this.getNaming() === `sql`) {
      if (this.currentSchemaStore)
        return this.currentSchemaStore;

      const result = await this.execute<{ '00001': string }>(`values (current schema)`);
      if (result.success && result.data.length > 0) {
        this.currentSchemaStore = result.data[0]['00001'];
        return this.currentSchemaStore;
      }
    }

    return this.options.libraries && this.options.libraries.length > 0 ?
      this.options.libraries[0] : `QGPL`;
  }

  getNaming(): NamingFormats | undefined {
    return this.options.naming;
  }

  async setSelfState(code: SelfValue) {
    await this.query<any>(`SET SYSIBMADM.SELFCODES = '${code}'`).execute();
    this.selfState = code;
  }

  getJobLog(): Promise<QueryResult<JobLogEntry>> {
    return this.query<JobLogEntry>(`select * from table(qsys2.joblog_info('*')) a`).execute();
  }

  async requestCancel(): Promise<boolean> {
    const instance = getInstance();
    const connection = instance.getConnection();

    // Note that this statement is run via the base extension since it has to be done on a job other than the one whose SQL is getting canceled
    await connection.runSQL(`CALL QSYS2.CANCEL_SQL('${this.id}')`);

    const [row] = await connection.runSQL(`select V_SQL_STMT_STATUS as STATUS from table(qsys2.get_job_info('${this.id}'))`) as { STATUS: string | null }[];
    return !(row && row.STATUS === `ACTIVE`);
  }
}
```

src/views/jobManager/editJob/systemTab.ts
```diff
@@ -120,10 +119,7 @@ export default function getSystemTab(options: JDBCOptions) {
     //   ],
     //   formatDescription(autoCommitText)
     // )
-    .addSelect(`database name`, `Database name`, [
-      {text: `System Base`, description: `*SYSBAS`, value: ``, selected: options["database name"] === ``},
-      ...Object.values(connection.getAllIAsps()).map(asp => ({text: asp.name, description: asp.name, value: asp.rdbName, selected: options["database name"] === asp.rdbName}))
-    ], formatDescription(dbNameText))
```
SanjulaGanepola, line 126: Why do we no longer fetch this and instead rely on the user to set it? From what I see in Code4i, this API is still available.

Is this related to https://github.com/codefori/vscode-db2i/issues/546 which was just opened? In which case, lets link that issue to this PR too.

src/views/results/index.ts
```diff
@@ -299,30 +299,31 @@ async function runHandler(options?: StatementInfo) {
...
     let possibleTitle = `SQL Results`;
     if (ref && ref.object.name) {
       possibleTitle = (ref.object.schema ? ref.object.schema + `.` : ``) + ref.object.name;
     }
 
-    if (statement.type === StatementType.Create || statement.type === StatementType.Alter) {
+    if (ref && (statement?.type === StatementType.Create || statement?.type === StatementType.Alter)) {
       const databaseObj =
-        statement.type === StatementType.Create && ref.createType.toUpperCase() === `schema`
+        statement.type === StatementType.Create && ref?.createType?.toUpperCase() === `schema`
```
SanjulaGanepola:
```suggestion
        statement.type === StatementType.Create && ref.createType?.toUpperCase() === `schema`
```

src/extension.ts
```diff
@@ -31,22 +30,27 @@ import { QueryHistory } from "./views/queryHistoryView";
...
+  sqlJob: (options?: JDBCOptions) => Promise<ExtendedSQLJob>
   sqlExamples: typeof SQLExamples,
   statement: typeof Statement,
   dbCache: typeof DbCache,
 }
 
-// this method is called when your extension is activated
-// your extension is activated the very first time the command is executed
+export namespace Db2foriOutput {
+  const outputChannel = vscode.window.createOutputChannel(`Db2 for i Server Component`, `json`);
```
SanjulaGanepola: Since there is no more server component, how about just calling it the extension name? The output channel isn't used for much right now, but maybe we can start to add more logging down the line :)

```suggestion
  const outputChannel = vscode.window.createOutputChannel(`Db2 for IBM i`, `json`);
```

src/views/jobManager/editJob/otherTab.ts
```diff
@@ -585,13 +573,13 @@ export default function getOtherTab(options: JDBCOptions) {
     .addInput(
       `key ring name`,
       `Key Ring Name`,
-      `Specifies the key ring class name used for SSL connections with the system. This property has no effect unless "secure" is set to true and a key ring password is set using the "key ring password" property.`,
+      `Specifies the key ring class name used for SSL connections with the system. This property has no effect unless "Use secure SQL connection" is enabled at the server connection level and a key ring password is set using the "key ring password" property.`,
```
SanjulaGanepola, line 576: What is `Use secure SQL connection` referring to now that the `secure` option was removed?

src/views/jobManager/editJob/systemTab.ts
```diff
@@ -120,10 +119,7 @@ export default function getSystemTab(options: JDBCOptions) {
...
+    .addInput(`database name`, `Database name`, undefined, { default: options["database name"] })
```
SanjulaGanepola: Description is missing

Image

```suggestion
    .addInput(`database name`, `Database name`, formatDescription(dbNameText), { default: options["database name"] })
```

src/extension.ts
```diff
@@ -31,22 +30,27 @@ import { QueryHistory } from "./views/queryHistoryView";
...
-  sqlJob: (options?: JDBCOptions) => OldSQLJob
+  sqlJob: (options?: JDBCOptions) => Promise<ExtendedSQLJob>
   sqlExamples: typeof SQLExamples,
   statement: typeof Statement,
   dbCache: typeof DbCache,
 }
 
-// this method is called when your extension is activated
-// your extension is activated the very first time the command is executed
+export namespace Db2foriOutput {
```
SanjulaGanepola, line 39: Shall we put this namespace into its own file?

src/connection/manager.ts
```diff
@@ -26,60 +26,53 @@ export class SQLJobManager {
...
-            options = savedConfig;
-          }
-        }
-
-        if (!options.libraries) {
-          options.libraries = [config.currentLibrary, ...config.libraryList.filter((item) => item != config.currentLibrary)];
+  async newJob(options?: JDBCOptions, name?: string) {
+    const connection = getInstance().getConnection()!;
+    const config = connection.getConfig();
+    const mapepire = await connection.getComponent<any>("mapepire");
```
SanjulaGanepola, line 32: This requires the import suggestion above too.

```suggestion
    const mapepire = await connection.getComponent<Mapepire>("mapepire");
```

src/connection/manager.ts
```diff
@@ -1,17 +1,17 @@
...
 import { getInstance } from "../base";
-import { askAboutNewJob, Config, onConnectOrServerInstall, osDetail } from "../config";
+import { askAboutNewJob, Config, osDetail } from "../config";
 import Configuration from "../configuration";
-import { SelfValue } from "../views/jobManager/selfCodes/nodes";
-import { ServerComponent, UpdateStatus } from "./serverComponent";
-import { OldSQLJob } from "./sqlJob";
 import { ConfigManager } from "../views/jobManager/ConfigManager";
+import { SelfValue } from "../views/jobManager/selfCodes/nodes";
+import { ExtendedSQLJob, extendSQLJob } from "./extendedSQLJob";
```
SanjulaGanepola, line 10:
```suggestion
import { ExtendedSQLJob, extendSQLJob } from "./extendedSQLJob";
import { Mapepire } from "@halcyontech/vscode-ibmi-types/api/components/mapepire";
```

src/views/jobManager/statusBar.ts
```diff
@@ -42,7 +41,7 @@ export async function updateStatusBar(options: {newJob?: boolean, canceling?: bo
           ``,
           `Configured user library list for job:`,
           ``,
-          ...job.options.libraries.map((lib, i) => `${i+1}. \`${lib}\``)
+          ...job.options?.libraries?.map((lib, i) => `${i+1}. \`${lib}\``) || 'none'
```
SanjulaGanepola: To avoid the `none` looking like this:
Image
```suggestion
          ...(job.options?.libraries?.map((lib, i) => `${i+1}. \`${lib}\``) || ['N/A'])
```

src/views/jobManager/editJob/systemTab.ts
```diff
@@ -1,6 +1,5 @@
 import { JDBCOptions } from "@ibm/mapepire-js/dist/src/types";
-import { getBase, getInstance, loadBase } from "../../../base";
-import { formatDescription } from ".";
+import { getBase, getInstance } from "../../../base";
```
SanjulaGanepola, line 2: This last import is required for a suggested further fix below.

```suggestion
import { getBase, getInstance } from "../../../base";
import { formatDescription } from ".";
```

src/extension.ts
```diff
@@ -31,22 +30,27 @@ import { QueryHistory } from "./views/queryHistoryView";
...
+  const outputChannel = vscode.window.createOutputChannel(`Db2 for i Server Component`, `json`);
+  export function writeOutput(jsonString?: string, show = false) {
```
SanjulaGanepola, line 41: From my understanding of this change, we won't be able to see logs like this anymore. Is this correct?

Image

src/language/sql/document.ts
```diff
@@ -63,7 +63,7 @@ export default class Document {
           break;
 
         case `statementType`:
-          currentStatementType = StatementTypeWord[upperValue];
+          currentStatementType = (upperValue ? StatementTypeWord.get(upperValue) : StatementType.Unknown) || StatementType.Unknown;
```
SanjulaGanepola, line 66: This same change needs to happen here (and I think these errors really need to get caught by the build):

Image

or else it will break completions support (and possibly other things):

Image

I tried the change:
```ts
this.type = (wordValue ? StatementTypeWord.get(wordValue) : StatementType.Unknown) || StatementType.Unknown;
```

But it actually seems like the completions are still broken:

Image

Also, invalid completions here too:

Image

SanjulaGanepola (changes requested):
Okay, I tried to do hopefully a thorough review of this (lots of changes!) and commented anything and everything I found 😅

Just wanted to point out that it looks like there are a whole lot of type errors still in several files. Are these unrelated to the changes made in this branch? Also not sure why these are not flagged by the build.

Image

SanjulaGanepola:
@forstie After the comments I just left are addressed, it would be great if you could test this out as well just to make sure nothing was broken.

forstie:
I'm ready to test it when given the go ahead.
It would help to have a brief summary of where to focus testing.

sebjulliand:
@SanjulaGanepola I tried to fix most of the type errors. It must be ok now (I put //@ts-ignore on some of them because these cases would need more changes).

src/connection/extendedSQLJob.ts
sebjulliand: The problem is we don't actually instantiate the SQLJob. Mapepire component from Code for i does. So now we're more like wrapping it.
I tried to keep it extended but it didn't go well.

src/connection/manager.ts
sebjulliand: I tried that...the build blew oup, saying it didn't find the type. I tried to change the TS and Webpack settings, to no avail.
So I used `any` and hardcoded "mapepire" name to make it work.

src/views/jobManager/editJob/otherTab.ts
sebjulliand: It's managed by the core, at the connection settings level.

src/views/jobManager/editJob/systemTab.ts
sebjulliand: That's been removed from the API. The `ASP_INFO` service that gave us this info cannot be run by any user in 7.6. So listing ASP is not possible anymore, hence the switch to a simple input here.

src/views/jobManager/editJob/otherTab.ts
sebjulliand: This, in the core connection settings.
image

SanjulaGanepola:
@sebjulliand Sounds good, let me know when this is ready for a re-review :)

src/views/jobManager/editJob/otherTab.ts
SanjulaGanepola: Ahh, thanks for the clarification

src/views/jobManager/editJob/systemTab.ts
SanjulaGanepola: Oh didn't know that. Is there any alternatives we can use to still fetch this @forstie? Or input box the way to go?

src/views/jobManager/editJob/otherTab.ts
SanjulaGanepola: Yup, completely missed that we added it here. Thanks again for clarification

src/connection/manager.ts
SanjulaGanepola: Hm interesting. When I briefly tried it yesterday on my end, nothing blew up 😅
But not a huge deal so we can do without it the issue persists

src/connection/components/checkStatement.ts
```diff
@@ -0,0 +1,91 @@
...
+
+export class CheckStatementComponent implements IBMiComponent {
+  static ID = "CheckStatementComponent";
+  private static readonly VERSION = 1;
+  static readonly FUNCTION_NAME = `CHKSTMNT${CheckStatementComponent.VERSION.toString().padStart(4, "0")}`;
+  private static readonly SIGNATURE = "7CDD7F08E0CE36EF271A81197C2D770CBBD3A3AA2E02D3217421995EA555A7F0";
+  private static readonly VALID_STATEMENT_LENGTH = 32740;
+  private static readonly TYPE = "PROCEDURE";
+
+  private getLibrary(connection: IBMi) {
```
SanjulaGanepola, line 14: Doesn't look like this is actually used. There are a couple instances in this file below where we are doing this logic. Shall we make use of this func?

src/connection/components/validateStatement.ts
```diff
@@ -0,0 +1,292 @@
...
+      name: ValidateStatementComponent.ID,
+      version: ValidateStatementComponent.VERSION,
+      signature: ValidateStatementComponent.SIGNATURE,
+    };
+  }
+
+  async getRemoteState(connection: IBMi, installDirectory: string): Promise<SecureComponentState> {
+    const remoteSignature = await this.getSQLRoutineSignature(
+      connection,
+      connection.getConfig().tempLibrary.toUpperCase(),
```
SanjulaGanepola: Can we use the `getLibrary` func

src/connection/components/validateStatement.ts
```diff
@@ -0,0 +1,292 @@
...
+    return (await connection.runSQL(
+      /* sql */`select HASH_SHA256(ROUTINEDEF) SIGNATURE from qsys2.sysroutines where routine_type = '${type}' and rtnschema = '${library}' and RTNNAME = '${name}' fetch first row only`
+    )).pop()?.SIGNATURE as string;   
+  }
+
+  async update(connection: IBMi, installDirectory: string): Promise<SecureComponentState> {
+    return connection.withTempDirectory(async tempDir => {
+      const tempSourcePath = getVSCodeTools()?.ensureFullPath(posix.join(tempDir, `sqlvalidator.sql`), connection?.getConfig()?.homeDirectory);
+      const srcPath = tempSourcePath ? tempSourcePath : '';
+      const library = connection?.getConfig()?.tempLibrary.toUpperCase() || `ILEDITOR`;
```
SanjulaGanepola: Can we use the `getLibrary` func

src/views/results/explain/doveTreeDecorationProvider.ts
```diff
@@ -51,16 +51,14 @@ export class DoveTreeDecorationProvider implements FileDecorationProvider {
                 } else {
                     color = nodeHighlights.getPriorityColor();
```
SanjulaGanepola, line 52: Image

src/views/results/explain/doveTreeDecorationProvider.ts
```diff
@@ -51,16 +51,14 @@ export class DoveTreeDecorationProvider implements FileDecorationProvider {
...
                     badge = String(nodeHighlights.getCount()); // The number of highlights set for the node
+                    //@ts-ignore
                     tooltip = "\n" + nodeHighlights.getNames().map(h => "🔥 " + Highlighting.Descriptions[Highlighting[h]]).join("\n");
                 }
                 return {
-                    color: color,
-                    badge: badge,
-                    tooltip: tooltip,
```
SanjulaGanepola, line 59: Why was the badge and tooltip dropped?

SanjulaGanepola (changes requested):
Not sure if these type errors are due to changes made in this PR, but thought I would just point them out in case we wanted to fix them here. If they are unrelated, we can maybe leave for a different PR:

Image

Did we ever find why these aren't flagged by the build?

src/language/sql/document.ts
SanjulaGanepola: @sebjulliand I am still hitting these issues with completions

julesyan:
@SanjulaGanepola resolved the two comments for the components
@sebjulliand the other comments are for your code

SanjulaGanepola:
> I'm ready to test it when given the go ahead.
> It would help to have a brief summary of where to focus testing.

@forstie For testing, this PR doesn't actually have any changes that actually impact functionality. Rather we are no longer installing and have our own mapepire logic in this extension, and instead re-using similar mapepire logic which was added in Code for IBM i. > 3.0.0. However, since this PR does touch a lot of files though, I just tested everything more generally to make sure nothing was broken. I am hoping you can help do the same.

Note: Content assist is not working properly right now, but everything else seems to be good. I have a comment above for that.

Note the VSIX can be found here: https://github.com/codefori/vscode-db2i/pull/539#issuecomment-4604488058
Also for testing, make sure you are on Code for IBM i 3.0.10.

src/views/results/explain/doveTreeDecorationProvider.ts
sebjulliand: My mistake

sebjulliand:
@SanjulaGanepola I fixed those types issue and the completion (hopefully).
I still don't know why the build doesn't complain about the TS issues...

SanjulaGanepola (approved):
Awesome, all looks good on my end

forstie (approved):
I'll test post release.

PR #552, opened 2026-06-12
Fix validate_statement to not rely on hardcoded library

### LINKED ISSUE #549: ValidateStatement warning message
**Describe the bug**
Every time I connect to my IBM i I get the warning message.
I tried to clear the temp library and the temp directory and reconnect, but I get the same warning.

**Screenshots**

Image

**Environment:**
 - Extension version 3.0.10
 - IBM i OS version 7.5

julesyan:
### Changes
Fixes #549, the validate statement component was using a hardcoded library in the source. We have removed the  qualifier so it uses the default library of the RUNSQLSTM job (which is already set properly) 

### How to test this PR
1. Connect to a system where the temp library is not ILEDITOR
2. Once connected edit an SQL file to trigger the syntax checker

### Checklist

* [x] have tested my change
* [x] Remove any/all `console.log`s I added

### DIFF
--- a/src/connection/components/validateStatement.ts
+++ b/src/connection/components/validateStatement.ts
@@ -79,7 +79,8 @@ export class ValidateStatementComponent implements IBMiComponent {
       const tempSourcePath = posix.join(tempDir, `sqlvalidator.sql`);
       const srcPath = tempSourcePath ? tempSourcePath : '';
       const library = this.getLibrary(connection);
-      await connection.getContent().writeStreamfileRaw(srcPath, this.getSource(library, ValidateStatementComponent.VERSION));
+      const oldLib = (await connection.runSQL('VALUES CURRENT PATH'))[0]['00001'] as string;
+      await connection.getContent().writeStreamfileRaw(srcPath, this.getSource(library, ValidateStatementComponent.VERSION, oldLib));
       const result = await connection.runCommand({
         command: `QSYS/RUNSQLSTM SRCSTMF('${tempSourcePath}') COMMIT(*NONE) NAMING(*SYS) DFTRDBCOL(${library})`,
         cwd: `/`,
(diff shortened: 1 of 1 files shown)


SanjulaGanepola:
> it uses the default library of the RUNSQLSTM job (which is already set properly)

If we don't qualify it, my understanding would be that it looks it up in the caller's library list or am I wrong about that?

julesyan:
I tested this with a new library where the only thing in my library list was QGPL and QTEMP, so it looks like its grabbing the correct function @forstie might be able to help more on this

sebjulliand:
I thinks it's going to be a fun one.
My assumption is it will depend on the `naming` option of the job.
- `*SYS`: it will look in the library list
- `*SQL`: it will look in the `SQL PATH`

Not entirely sure but from my past experience, it was something like that.
And we can't assume what the job configuration is going to be.

For now, it doesn't work (it says "CHKSTMNT0001 of type *N not found in ILEDITORX"):
image

Even though it exists:
image

SanjulaGanepola:
Why not make the library a parameter on the SQL function we are creating so when we calculate the hash value, it is not dependent on any hardcoded library name?

sebjulliand:
No, but I'm not sure that would work (assuming IN_LIBRARY is a parameter):
```sql
call IN_LIBRARY.CHKSTMNT(...)
```

Worth a try though!

julesyan:
Yea setting it as a param does not work, i get `[SQL0204] CHKSTMNT0001 in LIB type *N not found.`

julesyan:
Ok @sebjulliand if you could test now with your case. This worked on my local but im not sure if im missing a case

sebjulliand (approved):
Great job @julesyan , it works like a charm now!
Tested with `*SYS` and `*SQL` naming. No issues.

I just had to delete the function because of my previous test, but that won't happen to the users.

PR #556, opened 2026-06-29
Refactoring SQL joblog view

buzzia2001:
### Changes
With this PR, the view for WRKJOB has been updated to use the one from the FS extension.

### How to test this PR
1. Connect to a system and hover over the name of the database connection profile
image

2. A new tab containing WRKJOB will open
image

### Checklist

* [x] have tested my change


sebjulliand:
@buzzia2001 I like the idea, but there are two suggestions I would make:
- There is the SQL job from the core that could benefit from having a joblog viewer
  image
- In this case, the joblog webview should be moved to the core, so there is no circular dependency between Core and FS

What do you think ?

NicolasSchindler:
@sebjulliand @buzzia2001 and i are already on that :) I prepare a PR later for the Core. Also to include the JDBC configuration of the Mapepire job

sebjulliand:
> @sebjulliand @buzzia2001 and i are already on that :) I prepare a PR later for the Core. Also to include the JDBC configuration of the Mapepire job

Ok, cool!
Be careful about putting the JDBC options in the Core. The core's job is not meant to be overly configurable since it's used for the core's processes. It would be a shame if users could break it by misconfiguring it.

buzzia2001:
Ciao @sebjulliand @NicolasSchindler ,

I'll I'll move the WRKJOB view into the core extension, but I also have to move the frontend, so it'll take a little while.

sebjulliand:
@buzzia2001 sure, no rush 😄

buzzia2001:
@SanjulaGanepola and @sebjulliand 

I've updated this PR to use the new joblog command that's included in the core. I've also added the job ID to the tooltip; you can copy it just like on ACS if you want.

image

src/views/jobManager/jobManagerView.ts
```diff
@@ -89,21 +88,18 @@ export class JobManagerView implements TreeDataProvider<any> {
       }),
 
       vscode.commands.registerCommand(`vscode-db2i.jobManager.viewJobLog`, async (node?: SQLJobItem) => {
-        const id = node ? node.label as string : undefined;
```
SanjulaGanepola, line 92: Don't we still need this `node` logic as the action is also available inline in the SQL Job Manager

src/views/jobManager/statusBarColor.ts
```diff
@@ -0,0 +1,19 @@
+const HEX_COLOR = /^#([0-9a-f]{3}|[0-9a-f]{6})$/i;
+
+/** The colour picker has no empty state, so black is what it reports when no colour was ever picked. */
+const NO_COLOR = `#000000`;
+
+/**
+ * Normalizes a status bar colour into `#rrggbb`; returns undefined when it's empty, invalid or black,
+ * which leaves the status bar item with the colour of the current theme.
+ */
+export function parseStatusBarColor(color?: string) {
```
SanjulaGanepola: Later since this is duplicate code from Code4i, we should consider exporting it in `VscodeTools` in the base and then re-using that.

SanjulaGanepola (changes requested):

src/views/jobManager/statusBarColor.ts
SanjulaGanepola: This is not a big deal, so we can just do this later

buzzia2001:
@SanjulaGanepola updated as requested

SanjulaGanepola (approved):
Works great!

Before we do a release, we should make sure to do a bump of the Code4i types once that is released to get rid of:

Image

PR #560, opened 2026-07-17
Adjust auto commit and transaction isolation text

SanjulaGanepola:
### Changes
- Adjust transaction isolation level text based on https://www.ibm.com/docs/en/i/7.4.0?topic=transactions-transaction-isolation-levels
- Fix auto commit text


forstie:
Is it possible to adjust this slightly to show the IBM i specific alternative value?
I ask because I think more users know *CHG vs *UR, and so on.
image

SanjulaGanepola:
If I understand correctly, show both alternatives?

image

forstie (approved):
brilliant, thanks

PR #565, opened 2026-08-03
Call actions for procedures and functions

### LINKED ISSUE #221: Schemas - Generate an SQL statement for routines
I know we've got a nifty prompter / content assist, but it would also be nice to generate CALL statements for procedures from Schemas and Queries for functions from Schemas... similar to what Schemas does for Tables and Views.

For procedures, generate a CALL proc();
For scalar functions, generate a VALUES fun();
For table functions, generate SELECT  * from table(fun());

buzzia2001:
### Changes
This PR enables the generation of CALL or SELECT statements for procedures and functions (table and scalar).

### How to test this PR
From the SCHEMA BROWSER, select a procedure or a function (scalar or table), the “Generate Call Statement” option will appear. Click it. A new tab will open with the SQL code for querying or calling the function/procedure.
image

Procedure:
image

Scalar function:
image

Table function:
image

@forstie @ryan-moeller21 could you test it?

### Checklist
* [x] have tested my change

Closes #221

### DIFF
--- a/src/database/callable.ts
+++ b/src/database/callable.ts
@@ -3,6 +3,7 @@ import { JobManager } from "../config";
 import { SQLParm } from "../types";
 
 export type CallableType = "PROCEDURE"|"FUNCTION";
+export type FunctionType = "SCALAR"|"COLUMN"|"TABLE";
 export interface CallableRoutine {
   schema: string;
   name: string;
@@ -35,6 +36,21 @@ export default class Callable {
     if (result.length > 0) {
       routine.specificNames = result.map(row => row.SPECIFIC_NAME);
       return routine;
+    }
+
+    return;
+  }
+
+  static async getFunctionType(schema: string, specificName: string): Promise<FunctionType|undefined> {
+    const result = await JobManager.runSQL<{FUNCTION_TYPE: string}>(
+      `select FUNCTION_TYPE from qsys2.sysfuncs where SPECIFIC_SCHEMA = ? and SPECIFIC_NAME = ?`,
+      {parameters: [schema, specificName]}
+    );
+
+    switch (result[0]?.FUNCTION_TYPE?.trim()) {
+      case `S`: return `SCALAR`;
+      case `C`: return `COLUMN`;
+      case `T`: return `TABLE`;
     }
 
     return;
--- a/src/views/schemaBrowser/contributes.json
+++ b/src/views/schemaBrowser/contributes.json
@@ -51,6 +51,11 @@
         "category": "Db2 for i"
       },
       {
+        "command": "vscode-db2i.generateRoutineCall",
+        "title": "Generate call statement",
+        "category": "Db2 for i"
+      },
+      {
         "command": "vscode-db2i.getRelatedObjects",
         "title": "Get Related Objects",
         "category": "Db2 for i"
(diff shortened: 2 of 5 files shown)


src/database/callable.ts
```diff
@@ -40,6 +41,21 @@ export default class Callable {
...
+  static async getFunctionType(schema: string, specificName: string): Promise<FunctionType|undefined> {
+    const result = await JobManager.runSQL<{FUNCTION_TYPE: string}>(
+      `select FUNCTION_TYPE from qsys2.sysfuncs where SPECIFIC_SCHEMA = ? and SPECIFIC_NAME = ?`,
+      {parameters: [schema, specificName]}
+    );
+
+    switch (result[0]?.FUNCTION_TYPE?.trim()) {
+      case `S`: return `SCALAR`;
+      case `C`: return `COLUMN`;
+      case `T`: return `TABLE`;
```
ryan-moeller21, line 53: QSYS2.SYSFUNCS will return a blank if the entry corresponds to a procedure. Will this ever be run against a procedure, and should it handle that procedure nicely?

I see that the call to `getFunctionType` only occurs if `object.type === `function`` returns true, but I'm not familiar with the `SQLObject` class.

buzzia2001: Hi @ryan-moeller21, 
getFunctionType is called only when object type is function: 

image

If the function receives a different value, it returns null, which then becomes SCALAR as the default value.

src/views/schemaBrowser/statements.ts
```diff
@@ -1,4 +1,43 @@
 import Statement from "../../database/statement";
+import { prepareParamType } from "../../language/providers/logic/completion";
+import { SQLParm } from "../../types";
+
+export type RoutineInvocation = "PROCEDURE" | "SCALAR" | "TABLE";
+
+export function getRoutineCallStatement(schema: string, name: string, parms: SQLParm[], invocation: RoutineInvocation) {
+  const qualifiedName = `${Statement.delimName(schema)}.${Statement.delimName(name)}`;
```
SanjulaGanepola: Shouldn't we check if the current job is using sql or system naming and use the appropriate separator?

SanjulaGanepola (changes requested):

src/views/schemaBrowser/statements.ts
```diff
@@ -1,4 +1,43 @@
...
+  // The arguments of a table function sit one level deeper, inside the TABLE() clause
+  const call = `${qualifiedName}(${getArgumentList(parms, invocation === `TABLE` ? `  ` : ``)})`;
+
+  switch (invocation) {
+    case `PROCEDURE`: return `CALL ${call};`;
+    case `TABLE`: return `SELECT * FROM TABLE(${call}) as a;`;
```
SanjulaGanepola: For casing consistency
```suggestion
    case `TABLE`: return `SELECT * FROM TABLE(${call}) AS A;`;
```

src/views/schemaBrowser/contributes.json
```diff
@@ -50,6 +50,11 @@
         "title": "With Separate ALTER Statements",
         "category": "Db2 for i"
       },
+      {
+        "command": "vscode-db2i.generateRoutineCall",
+        "title": "Generate call statement",
```
SanjulaGanepola:
```suggestion
        "title": "Generate Call Statement",
```

SanjulaGanepola (changes requested):
@buzzia2001 Couple minor comments and then please merge. Nice work

PR #567, opened 2026-08-04
Add mti actions

### LINKED ISSUE #563: Add interactive menu for Maintained Temporary Indexes (MTIs) for tables in the schemas browser
It is possible today to view the MTIs built for tables in a certain library by right-clicking on a table and clicking the "Get MTIs" button. However, it's not possible to do anything with this information besides viewing the MTIs.

Image

It would be great to be able to interact with each MTI. For example, I would like to right-click and create a perm index that matches the definition of a given MTI. ACS can do this today (see below)

Image

Image

buzzia2001:
### Changes
This PR allows you to perform actions on MTIs. MTIs can be managed at the individual library or individual object level.
The supported actions are generating the SQL statement for the index or creating the index itself.

### How to test this PR
From schema browser, choose MTI actions by right-clicking on a library or a file:
image

Clicking on “MTI Actions” will display the MTIs based on the search criteria:
image

By clicking on the suggested MTI, you can choose between SQL generation and index generation. Here is the output if you choose to generate SQL:
image

If, on the other hand, you want to create the index directly, you will first be prompted to enter the index name, and then a pop-up will appear asking you to confirm execution of the statement:
image

@forstie @ryan-moeller21 could you test it?

### Checklist
* [x] have tested my change

Closes #563


buzzia2001:
@ryan-moeller21 but why don't we do the same thing for the Index Advisor? Right now, just like with MTI, it only shows a table with the list, but there's no way to interact with them

src/views/schemaBrowser/mti.ts
```diff
@@ -0,0 +1,200 @@
...
+const CREATE_INDEX = `Create Index...`;
+const SHOW_STATEMENT = `Show CREATE INDEX Statement`;
+
+function qualifiedTable(mti: MTIInfo): string {
+  return `${Statement.delimName(mti.TABLE_SCHEMA)}.${Statement.delimName(mti.TABLE_NAME)}`;
+}
+
+/** The key definition is already valid index key syntax, so it is used as-is */
+export function buildCreateIndexStatement(mti: MTIInfo, indexName: string): string {
+  const name = Statement.delimName(indexName, true);
```
ryan-moeller21, line 50: Super nit-picky here... The maximum length of an SQL object's name is 128 characters. If the table name is 128 characters long, this logic will build an index name that is greater than 128 characters long, making the CREATE INDEX statement invalid.

It looks like you anticipated this in the "create index" action (fantastic attention to detail!), but the "show CREATE INDEX statement" option generates an incorrect statement.

I like that you force the user to pick a valid index name in the "create index" action. We can't really do that in the "show statement" action, though. Should we truncate the name of the MTI so that a valid statement is created? Or, should we continue to generate an invalid index name, but maybe put in a comment above the CREATE INDEX statement that says something like "Warning: Invalid SQL name, index name length must be less than or equal to 128 characters."?

src/views/schemaBrowser/mti.ts
```diff
@@ -0,0 +1,200 @@
...
+
+  return [
+    `CREATE INDEX ${Statement.delimName(mti.TABLE_SCHEMA)}.${name}`,
+    `   ON ${qualifiedTable(mti)} (${mti.KEY_DEFINITION.trim()})`,
```
ryan-moeller21, line 55: If the MTI is a sparse index (see columns SPARSE and SPARSE_DEFINITION in the [MTI_INFO documentation](https://www.ibm.com/docs/en/i/7.6.0?topic=services-mti-info-table-function)), this CREATE INDEX statement will create an index over the entire set of rows, not a sparse set.

src/views/schemaBrowser/mti.ts
```diff
@@ -0,0 +1,200 @@
...
+    { modal: true, detail: statement },
+    `Create`
+  );
+
+  if (confirmation !== `Create`) return false;
+
+  try {
+    await vscode.window.withProgress(
+      { location: vscode.ProgressLocation.Notification, title: `Creating index ${name}...` },
+      () => JobManager.runSQL(statement)
```
ryan-moeller21: Is this CREATE INDEX statement run async, i.e. in a background job? 

If the user builds a very large index we don't want that CREATE INDEX statement to prevent the user from running other queries or creating other indexes. If the CREATE INDEX is "blocking", it would be good to fix. I know this would significantly complicate things (fetching success/error status of index build in a different job), so maybe this could be a future enhancement. If this is already handled, ignore this request 😄

ryan-moeller21 (commented):
Andrea, great work on this. Just a few small comments from me.

buzzia2001:
> Andrea, great work on this. Just a few small comments from me.

@ryan-moeller21 do you think we can also consider to do the same thing for index advisor?

ryan-moeller21:
> > Andrea, great work on this. Just a few small comments from me.
> 
> @ryan-moeller21 do you think we can also consider to do the same thing for index advisor?

Oops, sorry for missing your initial comment about the Index Advisor! This seems like a great next step, too.

buzzia2001:
Hi @ryan-moeller21,

1. I've managed sparse definition, unfortunatly in my demo partition I've nothing to test with it...
image

2. For the MTI names, I left a trailing padding of 9 characters in the format _MTIXXXXX, replace the various Xs with the MTI number padded with zeros

3. I've set it up so that if the user chooses to create the index, they submit a job in qusrnomax that creates the index... The job name is always C4ICRTIDX:
image

You can check that job directly in vscode:
image


<hr>

Could you please check point 1 please?

ryan-moeller21:
Nice work on changes 2 and 3.

The sparse changes look good to me! The result set shows that a sparse MTI exists on RMOELLER/MTI_TEST. I used your latest code to generate the CREATE INDEX statement, which works great. 

image

I'll approve the PR.

ryan-moeller21 (approved):

SanjulaGanepola:
Hey @buzzia2001. Thanks for working on this feature. I haven't yet fully reviewed the code / tested the changes, but my initial thoughts on the UI (related to the email you had sent as well):

1. The MTI info is definitely much better in a table instead of in a quick pick dropdown.
2. This PR introduces a new style of table. While it does align with the core and FS extension, the look doesn't align with all the other tables rendered via the database extension. In my opinion, I would say lets stick to the current style of table in the "Db2 for i" panel.
3. ~~Related to the point above, this PR renders the new table in a custom webview in the main editor space. I would honestly prefer to still have this in the bottom panel instead of the editor because it takes up too much space and also consistency. This is another reason I would suggest we stick to the current UI format.~~ (Edit: I take this back. I guess it depends on what is being displayed and the use case)
4. For actions on these MTIs, rather than buttons I would suggest right-click actions on a row make the most sense. We can definitely still add those in the current UI like so:
image

SanjulaGanepola:
@sebjulliand Curious to hear ur thoughts on this

SanjulaGanepola:
Another option, we make move to use `vscode-elements` for the table: https://vscode-elements.github.io/components/table/

buzzia2001:
I think my suggestion to put it in the middle (aside from keeping the UI consistent with the other extensions) was also due to the potential amount of information I might find there... hypothetically, I could have thousands of lines, and having to scroll through them all down there is, from my perspective, a bit “meh”... Also, the table below doesn’t have a search function, which means that if I have an MTI on a specific table and it’s the last one in the list, I have to scroll all the way to the bottom, which is absolutely “meh,” especially when I might have thousands of rows… 

Although, as I told Sanjula, I think we should do it like in ACS: on one hand, a view for query results at the bottom is fine; on the other hand, there are other features we’d like to add to the extension, such as MTI, Index Advisor with actions, etc., which are also full-screen in ACS... 

However, if the plan is to try updating the results table component to include a search (or filters) in that case as well, I think it could work... @SanjulaGanepola imho we should put this pr on hold until we test new component and rebase afther that, what do u think?

buzzia2001:
Hi @SanjulaGanepola @ryan-moeller21 @forstie 

I've pushed a new version following your requirements:
image

My idea is to use this space in other prs for:
- index advisor
- full self log (maybe filtered by user, but not only for the current job)
- maybe dbmon/placache content

Any feedback for me?

buzzia2001:
@SanjulaGanepola I also updated the colors to match those of the core:

image
image

SanjulaGanepola:
@buzzia2001 I like the look of it!

One thing though, what is the need for this new `DBA` view when the content is essentially a result set that could go into the current result sets view? What do you think about having a single view always for result sets, but we add an action here:

image

called `Move into Editor Area` that will pop it open into a webview into the editor. I think this is much better than having another view in the bottom panel especially when the content is just a table like the other view. This action gives the user the flexibility to then view multiple things at the same time which is what I believe you intended?

SanjulaGanepola:
It should probably be fairly easy to implement since we do this same thing here already with the run action in the editor:

image

We are essentially just letting the user move the result from the panel to the editor if they need.

buzzia2001:
@SanjulaGanepola i've made the change as requested

image
image

buzzia2001:
@sebjulliand @SanjulaGanepola @forstie @ryan-moeller21 

As requested, I've moved everything to a single component—both for result sets coming from SQL statements and those coming from actions, such as MTIs. Now that the query results have been moved to the new component, it's possible to click on the columns to sort the results or use the search bar to search within the result set; this encapsulates the user's query within another query handled by the component.

I'm attaching the compiled version of the extension. I would appreciate it if you could test it, since it involves a fairly significant change to the logic behind how results are handled.

[vscode-db2i-2.0.3-dev.0.vsix.zip](https://github.com/user-attachments/files/32442165/vscode-db2i-2.0.3-dev.0.vsix.zip)

image

For MTIs and, more generally, actions related to this component, I decided to place a button at the beginning of the line that allows you to interact with it; in the case of MTIs, this button lets you create the index or display the statement.

image

SanjulaGanepola:
Great work! I haven't reviewed the code yet, but did do some testing and noticed some issues and thought of some ideas.

### Issues

1. Sometimes after running a query, the `Refresh` button is always disabled:
image

2. Similar to above, sometimes the `Clear Results` is always disabled:
image

3. When the `Refresh` does work, it is appending rows:
image

3. When I run a query in the editor, I see that it shows the result set in the panel at the bottom. If I then re-run the same query, using `ctrl+r` on that same statement in the editor, it does not re-run it. This used to work.

5. For the MTI actions, I see the loaded row count is in the bottom left. But for a query it is in the bottom right. I think we should be consistent (I prefer the bottom left since that is where it was before). Can we also show the same style of message in both cases where we show how long the query took, what job, and the state?

image

image

6. For the `x` in the search bar, are we able to use the `close` codicon so it fits the theme as it is currently very hard to see?
image

7. Whenever I use the column width adjustment to drag a column to be wider, it also applies the sort by on that column which I don't want.
image

8. The lazy loading scrolling does not appear to work for me in some cases. It says "More available", but I cannot scroll. This seems to happen when the initial set of rows does not reach the bottom of the panel. This is reproducible by significantly lowering the `vscode-db2i.resultsets.rowsToFetch` setting to like 5.
image

### Ideas

1. Similar to the FS Extension, would it be possible for the `Search...` bar to highlight the match?

2. This one is not specific to this PR as I believe this was the same behavior as before, but I noticed that the 3 view actions we have in the panel are not available when the result set is moved to another view. Do you think this can be fixed in this PR?
image

3. Can we move the `Move this table into the editor` button to be with the others?
image

5. Can we have the `Move this table into the editor` also visible for all result sets too?

6. Now that we also display the query in the panel, what do you think about also adding a `Copy SQL` which will copy the query similar to ACS?

8. Since the filter and search actually change the query that is run, should we make it more visible to the user that this is the case since we don't actually update the query that is displayed at the top? My thought is can we show an icon to the left of the search bar when a modification (sort by or filter) is applied to indicate this? Then when we hover over this icon it can show exactly what the modification is?

image

10. Rather than have the `...` for `Actions`, can we support right-clicking the row? This is supported via `webview/context` as described [here](https://code.visualstudio.com/api/extension-guides/webview#context-menus):
image

buzzia2001:
@SanjulaGanepola I think I've addressed all the points you highlighted... I HOPE SO 😂😂

buzzia2001:
@SanjulaGanepola @ryan-moeller21 @forstie I've also updated index advisor to use same view as Work with MTIs:

image
image

SanjulaGanepola:
**1. Noticed this error:**

image

to avoid:

image

When I temporarily added a fix I see:

image

we need some kind of response message like before:

image

**2. `Clear Results` when used in the editor area leaves a empty webview. I would say we should just close it as it will never be re-used:**

image

**3. I noticed the enablement of view actions is not always correct.**

image

For this maybe we always have to leave this actions enabled and then when they are used, do the check for if they can be used?

**4. I noticed an odd issue where I get a warning message and the `null` value loses its yellow highlighting when I run an update statement, then click on a cell with `null`, and then click on a column heading adjuster to adjust the width:**

image

**5. In terms of the sort and filter hover, I was actually thinking it would be better if an icon to the left of the search bar would appear rather than just highlighting the existing search icon (this is so that it is more obvious as I didn't really notice the highlight at first or really think to hover on it).**. 

image

Also a normal `info` icon or something else might work to better indicate this?

**6. Right now when clicking on a column to sort by it, it swaps between ascending and descending. What if the user wants to remove this sort? Can the third click in the cycle remove it?**

**7. I noticed some inconsistencies with the filter and sort hover where it wasn't visible when working with MTIs and advised indexes, but worked for other result sets**

image

**8. Before when fetching advised indexes or MTIs, it would show that it is running the query in the panel. We should keep this behavior as it wasn't obvious that it was still running the query. I noticed there was a status bar item, but imo it can be missed easily if status bar is full (which is why i tend to add things to the status bar)**

image

We should have this same behavior for MTIs, Advised Indexes, and running really any query. 🚀🚀🚀

**8. When there were no advised indexes or MTIs found, we would previously still update the panel at the bottom. Now we only get an info notification. The reason I would say the previous behavior is better is because even if it didn't yield any results, we can still use the other actions to refresh, copy sql. With the current UI, we don't have the ability to do either.**

Before:

image

After:

image

**9. Do you think it would also be useful to just give an option to just run in the current job?**

image

**10. When they submit a job, I guess we have no way to know if the index was created and when? Can we add a button to view the log? Do we have the qualified job name?**

image

**11. When I run a query which does not yield any results, the footer is missing:**

image

**12. I was testing the column sorting feature on different column types and noticed it failed here. Are there other column types we should check for?**

image

PR #570, opened 2026-08-25
Fix parsing of host variables in obfuscated procedures and functions

SanjulaGanepola:
### Changes
WRAPPED statements have an obfuscated body that may contain colons as part of the encoding. Previously running this procedure/function like this would incorrectly detect host variables which would also cause a truncation issue.

https://www.ibm.com/docs/en/i/7.6.0?topic=functions-wrap

### How to test this PR
1. Generate an obfuscated function:
```sql
VALUES SYSIBMADM.WRAP('CREATE FUNCTION salary(wage DECFLOAT) RETURNS DECFLOAT 
                                RETURN wage * 40 * 52');
```
2. Run the function it generates:
```sql
CREATE FUNCTION SALARY ( WAGE DECFLOAT )  WRAPPED QSQ07040 aacxW8plW8VzG8pHG8VvG8Fv68Vn68Fbl8pJY8pJ1qpdW8pdW8pdW9pjaqebaqebakZuBeEd5uFkq8KJqqYsjm:1m0gyYOktd1gfJbJ0J8Pq1JRTtWhZ7eZZ7e4aa
```
3. Observe that the statement ran successfully.

### Checklist
* [x] have tested my change
* [x] have created one or more test cases

### DIFF
--- a/src/language/sql/statement.ts
+++ b/src/language/sql/statement.ts
@@ -660,6 +660,11 @@ export default class Statement {
 	 * DECLARE .. CURSOR FOR
 	 */
 	getEmbeddedStatementAreas() {
+		// WRAPPED statements have an obfuscated body that may contain colons as part of the encoding so avoid parsing them as host variables
+		if (this.tokens.some(t => tokenIs(t, `word`, `WRAPPED`))) {
+			return [];
+		}
+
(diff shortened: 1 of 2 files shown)


forstie (approved):
works great

PR #574, opened 2026-09-15
fix: preserve CREATE context when parsing CASE THEN expressions

### LINKED ISSUE #453: Invalid SQL statement selection and invalid syntax errors when `THEN` is used
Attempting to run the following statement fails as the selection range becomes invalid
Image

```sql
CREATE MASK mask_ccnum
  ON customers
  FOR COLUMN cc_number
  RETURN
    CASE
      WHEN cc_number IS NULL THEN NULL
      ELSE
        'XXXX-XXXX-XXXX-' || SUBSTR(cc_number, CHAR_LENGTH(cc_number) - 3, 4)
    END
  ENABLE;
```

Balrocj:
Fixes #453

## Problem

When running a `CREATE MASK` statement with the cursor on `THEN` inside a
`CASE` expression, the statement selector could incorrectly split the
statement at `THEN`.

This happens because `FOR COLUMN` was recognized as a `FOR` statement type,
overwriting the outer `CREATE` context. As a result, Db2 received an incomplete
fragment and returned `SQL0104`.

## Solution

Keep the current `CREATE` statement context when parsing a `FOR` token. This
prevents `THEN` within a `CASE` expression from being treated as the start of a
separate statement.

Before
image

After
image

## Validation

- Added a regression test verifying that a cursor on `THEN` selects the entire
  `CREATE MASK` statement.
- `npm test -- --reporter=dot` passes: 171 tests.
- `npm run webpack` completes successfully.
- Manual validation changed the server response from `SQL0104` to `SQL0552
  Not authorized to CREATE MASK`, confirming that Db2 receives the complete statement. The remaining authorization error is expected for the test user.


buzzia2001 (approved):
Thank you

